PR #9, opened 2017-11-01
Fix style for datetime picker

### LINKED ISSUE #6: Datetime pickers broken
Upstream issue is tempusdominus/bootstrap-4#18

youngbob:
Workaround for tempusdominus/bootstrap-4#18
Resolves #6


youngbob:
I changed the order of app.css and vendor.css so the vendor style could be overridden. I didn't notice any other issues from the change, since the only thing included in vendor.css at the moment seems to be the datetime picker.
It also seems like boostrap.[s]css should be included in vendor.css instead of app.css, but that isn't really necessary for this change.

cdubz:
Thanks, @youngbob! I suspect the swapping of `app`/`venodor.css` is fine, but I will do a bit of manual testing before I merge this.

I believe I ended up jamming the vendor stuff in to `app.css` because of issues with gulp not compiling the app styles without bootstrap variables available. Likely I've just done something weird with my gulp setup. I'll revisit that config at some point...

PR #10, opened 2017-11-01
Add validation for model durations

youngbob:
It was previously possible to accidentally enter obviously wrong values
for start/end dates. Add some basic validation to catch input errors
when creating or editing various models (when not using a timer).

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -1,9 +1,28 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from datetime import timedelta
+
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.template.defaultfilters import slugify
 from django.utils import timezone
+from django.utils.timesince import timesince
+
+
+def validate_duration(model, max_duration=timedelta(hours=24)):
+    """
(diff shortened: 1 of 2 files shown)


cdubz:
Nice! I actually have a few unpushed commits that establish validation for some forms, but I have not done anything with duration yet. Let me get those pushed and then we'll need to do a bit of merge work here.

cdubz:
@youngbob do you have any preference for doing a rebase vs. me resolving and doing a squash and merge? I have mostly used rebase myself and just tried the GH merge conflict web UI on your other PR. It's odd, hah.

youngbob:
I'm in the middle of rebasing and removing my changes that conflict with yours around the form errors.

youngbob:
So, apparently I did this a little differently from your other validation validation work. I think it makes more sense to do this kind of checking in the model validation (which is called by ModelForm).

cdubz:
Yep, I was just looking at that and I agree with your method. I'm still pretty new to Django and working my way around the form validation stuff was a little rocky (:

cdubz:
Thanks again! 👍

PR #19, opened 2017-11-18
Add ability to upload picture of child with thumbnailing capabilities.

overshard:
Another pull request for remote storage coming up next!


overshard:
Note, I can't seem to generate a `pipenv lock` / `Pipfile.lock` so I didn't include that here, you're going to need to generate that.

PR #24, opened 2017-12-01
Move ALLOW_UPLOADS to BABY_BUDDY dict

overshard:
This should work though I haven't tested as I'm running into issues with the use of dotenv python settings at the moment. You don't have to merge this yet just going ahead and making it a PR.


overshard:
This should be good to go now, it also fixes the issue of a manifest file needing to be present for development which is a minor bug for dev testing without first running the app.

PR #40, opened 2018-05-04
manual build corrections in readme.md

novski:
I figured out some issues in manual build, hope this helps others to build manualy faster...
Session variable is needed on raspberrypi 3+. Allthroug im not shure if it's the best way to go...

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,11 +185,10 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
 1. Default python3 to python and pip3 to pip for this session
 
         alias python=python3
-        alias pip=pip3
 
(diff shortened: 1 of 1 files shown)


cdubz:
Thanks for the PR.

Interesting note - the `sudo -H pip install pipenv` does not work when I tested walking through this on a Pi with a fresh install of raspbian strecth. I wonder if the alias does not work in python2 pip isn't installed?

Anyway, could you push another commit removing the `alias pip=pip3` line from step 2 and changing step 3 to `sudo -H pip3 install pipenv`? I think that is the only place that `pip3` is needed. I'll accept this PR with those changes.

Issue #43, opened 2018-05-05
ARM (i.e. Raspberry Pi) install problematic


cdubz:
#41 and #42 seem to indicate some issues with doing a manual install on ARM devices. I'm going to work through testing this on a Model B from Raspbian Stretch Lite to see if things can be smoothed out.

cdubz:
I am getting random kernel panics and segfaults with one of my Model B systems, but the other is getting closer. Perhaps the first one has memory or disk corruption.

Anyway, on my first run through I got an sha256 error similar to #41 but with a different package. I have found some random clues here and there about similar issues using pipenv's `Pipfile.lock` file between operating systems (e.g. bennuttall/piwheels#47; pypa/pipenv#1191) so I'm not sure there is much to do be done here (other than maybe advising `pipenv install --skip-lock` for the Pi specifically).

I am also testing running `pipenv lock` first, but that seems to run in to issues with the `psycopg-binary` requirement. I think maybe it doesn't have a binary for arm and fails to build from source because the system lacks some pg dev packages.

@novski Did you ever get things up and running on your Pi?

cdubz:
Just adding some additional notes from my testing - 

With my Pi, `pipenv install [...]` is consistently failing to build numpy, which is a dependency of pandas. Even if I rebuild the lock file on the Pi or use `--skip-lock` (which makes sense). The error I get is:

```
[...]

PR #45, opened 2018-05-07
tackle the manual a bit.

novski:
think it was because of the pipenv shell command that i used now that it worked so i add it to the manual... Also corrected a vim to editor line and split up the flow of point 12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
     **Note: Find the location of the pipenv virtual environment for the
     `virtualenv` parameter with the command `pipenv --venv`.**
 
-    Symlink it and restart:
+1. Symlink config and restart uWSGI:
 
         sudo ln -s /etc/uwsgi/apps-available/babybuddy.ini /etc/uwsgi/apps-enabled/babybuddy.ini
         sudo service uwsgi restart
--- a/README.md
+++ b/README.md
@@ -258,8 +258,8 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
     **Note: Find the location of the pipenv virtual environment for the
     `virtualenv` parameter with the command `pipenv --venv`.**
 
-    sudo ln -s /etc/uwsgi/apps-available/babybuddy.ini /etc/uwsgi/apps-enabled/babybuddy.ini
-    sudo service uwsgi restart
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,6 +210,7 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
 1. Initiate the Python environment
 
         pipenv install --three --dev
+        pipenv shell
```
cdubz: I don't think this is necessary here. If you take a look at your own log in #43, note that `pipenv shell` is never used. It is the other pieces that got you over the hump, I believe. Adding this doesn't actually *hurt* anything, but I would prefer you remove it.

README.md
```diff
@@ -256,12 +255,17 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
     See the [uWSGI documentation](http://uwsgi-docs.readthedocs.io/en/latest/)
     for more advanced configuration details.
 
-    **Note: Find the location of the pipenv virtual environment for the 
+    **Note: Find the location of the pipenv virtual environment for the
     `virtualenv` parameter with the command `pipenv --venv`.**
 
+    Symlink it and restart:
```
cdubz: I like this idea! It does make the flow of the steps much cleaner. But, please change this to treat the "Symlink it and restart" commands as a new step. I think that will make it even cleaner.

README.md
```diff
@@ -256,12 +255,17 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
...
+
+        sudo ln -s /etc/uwsgi/apps-available/babybuddy.ini /etc/uwsgi/apps-enabled/babybuddy.ini
+        sudo service uwsgi restart
+
 1. Create and configure the nginx server
```
cdubz: Please also separate these steps in the same way you did for the nginx config steps. Again, great idea!

README.md
novski: i assume it is the bart that make it work because  a frend told me on debian systems make differences when using sudo to user installations i think the environement keeps it in space.

README.md
```diff
@@ -180,7 +180,7 @@ Python 3.x, nginx, uwsgi and sqlite and should be sufficient for a few users
 
 1. Install Python 3.x, pip, nginx and uwsgi
 
-        sudo apt-get install python3 python3-pip nginx uwsgi uwsgi-plugin-python3 git
+        sudo apt-get install python3 python3-pip nginx uwsgi uwsgi-plugin-python3 git libopenjp2-7-dev
```
novski: i think i forgot to add that dependency

PR #51, opened 2018-07-16
fix icon

novski:
the icon did not show.

### DIFF
--- a/babybuddy/templates/babybuddy/nav-dropdown.html
+++ b/babybuddy/templates/babybuddy/nav-dropdown.html
@@ -4,7 +4,7 @@
 {% block nav %}
     <nav class="navbar navbar-expand-md navbar-dark bg-dark fixed-top">
         <a class="navbar-brand" href="/">
-            <img src="{% static "babybuddy/logo/icon-brand.png" %}" width="30" height="30" class="d-inline-block align-top" alt="">
+            <img src="{% static "babybuddy/root/icon-brand.png" %}" width="30" height="30" class="d-inline-block align-top" alt="">
             <span class="text-primary">Baby</span> Buddy
         </a>
 


cdubz:
Are you sure this fixed an issue for you? This PR affects the "brand" logo that appears in the upper left corner. The file is in `babybuddy/static_src/logo` so `{% static "babybuddy/logo/icon-brand.png" %}` is the correct code. Applying this PR breaks the brand logo.

novski:
well i was missing the logo after pulling the last commit and figured out that changeing from /logo/ to /root/ helps me get the logo back. This still happens when i change that line back and check with safari and firefox on MacOS with both browsers the same way. Neverless checking the file location shows icon-brand.png in /logo/icon-brand.png and /root/ does not list that file. So i checked for the definition of static and found a 'static' and a 'static_src' folder in my /var/www/babybuddy/public/babybuddy/ in that only 'static' folder there is no logo folder. Only a root where in i find a icon-brand.png. So ther must be a migrating conflict somewere. I still don't understand fully the gulp management so im not able to debug further. At least not in the next few days.

cdubz:
Hmmm. Yes maybe it's possible that I made the change without a proper migration. When you get a chance, could you check for the file in your deployed `static` folder? The png file should be in `/static/babybuddy/logo/`. If that is missing, I think you could resolve this by pulling the latest code and running `gulp build` and then `gulp collectstatic` (and maybe `gulp migrate` to be safe).

I actually want to start checking in the collect static files so people can skip this step or only rebuild them if they want to customize.

novski:
no there is no logo folder in static. I made runthrough to show you how it looks like now. 
bildschirmfoto 2018-07-19 um 19 54 17
I see now that also in https://github.com/cdubz/babybuddy/blob/master/babybuddy/settings/base.py line 134 the STATIC_URL = '/static/' -> should that not be /static_src/ ?

novski:
i stumbled over an double "static" declaration. Don't know if its relevant but whant to let you know its double here: public/etc/nginx.conf

cdubz:
The `STATIC_URL` setting is correct. Files are copied from the various `static_src` folders during the `collectstatic` operation.

I think you are right that there was some point when I changed this without a migration path. The steps I mentioned above *should* resolve this for you (but maybe backup files first, hah):

1. `git pull origin master`
1. `gulp build`
1. `gulp collectstatic`
1. `gulp migrate`

I'm going to close this as the change is not needed, but feel free to follow-up if you need more support (perhaps on Gitter for easier back and forth).

PR #58, opened 2018-09-05
Modify app manifest to fix iOS link issue. Fixes #57.

### LINKED ISSUE #57: Safari pinned iOS App links to Home / Permission denied
Visiting my Babybuddy with safari on iOS and pinning it to the Homescreen of my iPhone works. Taping on that nice babybudy icon on my Homescreen leads me to a page with a breadcrumb: Home / Permissions Denied. Switching back to Safari im still loged in. In menu there is a AnnonymousUser loged in that if i press logout leads to a switch over to Safari where i land on the login page.
Did someome ever noticed that?

HorizonXP:
This will need to be tested on Android to ensure it doesn't break anything there, but I don't envision that it does. Moreover, the simpler fix could be to ensure the web server provides `site.webmanifest` as a `Content-Type: application/json` MIME type.

### DIFF
--- a/babybuddy/static_src/root/site.webmanifest.json
+++ b/babybuddy/static_src/root/site.webmanifest.json
@@ -13,6 +13,9 @@
             "type": "image/png"
         }
     ],
+    "scope": "/",
+    "lang": "en-US",
+    "start_url": "/",
     "theme_color": "#ffffff",
     "background_color": "#ffffff",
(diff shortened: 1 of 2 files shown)


cdubz:
Interesting. Do you have any reference info for why the json extension change helps here? And have you tested this with Safari?

PR #74, opened 2019-07-08
Add dashboard sort clauses: first name and id

burkemw3:
I have seen the dashboard sorting be inconsistent. Last name then
first name then id should produce a consistent and predictable
sort order.


burkemw3:
Linting error fixed, please have another look :)

cdubz:
Thanks! 👍

PR #81, opened 2019-10-04
A couple minor changes

paulcalabro:
- Added support for persisting static content using the advice from https://github.com/cdubz/babybuddy/issues/46#issuecomment-394020997. Thanks @djmwj !
- Fixed a couple yamllint warning messages.
- Updated the version of the of the Compose file format to the latest version.
- Set some values explicitly to match `docker-compose config` output.


cdubz:
Thanks!

PR #83, opened 2019-10-27
Improve german tummy time translation

irmela:
I think "Bauchzeit" is a more accurate translation than "Spiel-Zeit".


cdubz:
Appears correct as far as I can tell 🤷‍♂ Thanks!

Issue #113, opened 2020-02-09
Allow application time zone to be set from UI


cdubz:
Currently the `TIME_ZONE` environment variable must be set for the application time zone. It would be more convenient if this could be changed from the frontend. There is currently no "global configuration"  for the application so this would need to be added.

cdubz:
Closed by #118.

PR #118, opened 2020-02-14
Allow application time zone to be set from UI

cdubz:
This PR enables the application timezone to be set per use, falling back on `settings.TIME_ZONE`.


cdubz:
Closes #113.

PR #132, opened 2020-03-13
Add API PATCH and DELETE methods

### LINKED ISSUE #126: Add PATCH support to API
As discussed in #125, Baby Buddy's API does not currently support partial (`PATCH`) updates to entries.

Django REST Framework appears to support this, though I have not looked at the particulars yet -- https://www.django-rest-framework.org/api-guide/serializers/#partial-updates

cdubz:
Closes #126.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ guess work.
   - [`GET` Method](#get-method)
   - [`OPTIONS` Method](#options-method)
   - [`POST` Method](#post-method)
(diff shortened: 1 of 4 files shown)


cdubz:
There are a couple models or serializers that will need to be updated. Tests are currently failing.

cdubz:
Aaand looks like tests are failing anyway because of the Python versions they use. Will come back to that as well...

PR #133, opened 2020-03-14
[WIP] Update serializers to accept a "timer" argument

### LINKED ISSUE #131: [Feature Request] API Endpoints to end timers and associate that timer with a task
## Basic functionality needed:
API endpoint for ending running timers, possibly via a PUT/PATCH request to the existing timer endpoint (#126).
Then either as a separate endpoint you should be able to associate a timer with a task like it's possible via the web interface. Either as a new endpoint, or by extending the existing POST endpoint to accept a timer id instead of a start & end parameter

## Possible use cases
I would very much like to integrate this app with my [Home Assistant](https://www.home-assistant.io/) instance. This would make it easy for me to set up physical buttons around the bed and the changing table that would post data to Baby Buddy.

I am well versed enough in Python and the Home Assistant code base, that creating a baby buddy component for Home Assistant would be trivial. But with the current API limitations it is severely limited in its use case because of the lack of timer management via the Baby Buddy API

## Why am i not just creating a pull request for this.
To be honest, i am not at all familiar with either Django as a whole or Django Rest Framework specifically. And while i did attempt to add basic PUT/PATCH functionality to the timer endpoint, the code quickly became ugly and unwieldy due to my unfamiliarity with the Django structure.

So now i am hoping that in due time somebody with the proper knowledge will be able to add these features.

cdubz:
Initial implementation for "timer" argument support in the API for models with duration (Feeding, Sleep, Tummy Time). Still needed:

- [x] New tests for all three models.
- [x] Custom information in the `OPTIONS` response.
- [x] Documentation updates.

Closes #131.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 from copy import deepcopy
 from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
 
 from django.contrib.auth.models import User
+from django.utils import timezone
 
 from core import models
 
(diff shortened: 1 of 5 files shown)


api/serializers.py
```diff
@@ -28,6 +30,62 @@ def validate(self, attrs):
...
+        if 'timer' in self.initial_data:
+            try:
+                timer = models.Timer.objects.get(pk=self.initial_data['timer'])
+            except models.Timer.DoesNotExist:
+                raise ValidationError({'timer': ['Timer does not exist.']})
+            if timer.end:
+                end = timer.end
+            else:
+                end = timezone.now()
+            if 'child' not in attrs and timer.child:
```
cdubz: Potentially this should not check for `child` in `attrs` and should just hammer over it as `start` and `end` do (it the Timer has a related Child).

PR #135, opened 2020-03-15
add enhanced feedings statistics

BenjaminHae:
Add statistics for the last 3 days / 2 weeks
easily configurable

### DIFF
--- a/dashboard/templatetags/cards.py
+++ b/dashboard/templatetags/cards.py
@@ -5,6 +5,8 @@ from django.db.models.functions import TruncDate
 from django.utils import timezone
 from django.utils.translation import gettext as _
 
+from datetime import date, datetime, time
+
 from core import models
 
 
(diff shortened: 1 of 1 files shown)


BenjaminHae:
I'm not sure, where the tests for this part are located.

dashboard/templatetags/cards.py
```diff
@@ -235,22 +238,40 @@ def _feeding_statistics(child):
...
+            }, 
+            {
+                'start': timezone.now()-timezone.timedelta(weeks=2), 
+                'title': _('Feeding frequency') + " 2 "+_("weeks")
+            },
+            {
+                'start': timezone.make_aware(datetime.combine(date.min, time(0,0)) + timezone.timedelta(days=1) ),
+                'title': _('Feeding frequency')
+            }
+        ]
```
cdubz: There is no need to concat the translated strings here. "Feeding frequency" is already a translation string, but "days" and "weeks" are not. Anyway, I'd prefer to have these strings as "Feeding frequency (past 3 days)" and "Feeding frequency (past 2 weeks)".

BenjaminHae:
Thank you for your input. Took me some time to setup gulp, but now I like it :)

I think I've fixed everything.

cdubz:
Yeah gulp is a bit of beast :hurtrealbad:

Thanks again for the contribution!

PR #137, opened 2020-03-24
Hide cards in dashboard without information

BenjaminHae:


cdubz:
@BenjaminHae thanks for the contribution!

Have you tested this with an empty DB (a child with no data)? I'm getting the following when I try to go to the dashboard for a child with nothing --

image

BenjaminHae:
> Have you tested this with an empty DB (a child with no data)? I'm getting the following when I try to go to the dashboard for a child with nothing --

Thank you, I did not think of that.

I've now tested it with a clean database and then created some data - seems to work.
Additionally when there is no data, there is now a small indicator that there is no data yet.

cdubz:
Thanks for the update!

I have to admit, I don't necessarily like this behavior. I'd like to see it controllable via a user setting to "hide empty dashboard cards" defaulting to the current behavior (of showing empty cards). Settings are in `babybuddy/models.py` and are available in templates via the `user.settings` global object. A migration will also be needed. Let me know if you'd like to do this work or I can merge this in and then tack it on.

BenjaminHae:
I'd like to give it a try, however it may take a while until I have time to do it.

jcgoette:
Glad to see this PR. Any movement? Can I help in any way?

cdubz:
@BenjaminHae do you still have any intention to come back around to this?

BenjaminHae:
Sorry, I could not find the time. 
@jcgoette if you could implement the change @cdubz requested, that would be nice.

> I have to admit, I don't necessarily like this behavior. I'd like to see it controllable via a user setting to "hide empty dashboard cards" defaulting to the current behavior (of showing empty cards). Settings are in babybuddy/models.py and are available in templates via the user.settings global object. A migration will also be needed. Let me know if you'd like to do this work or I can merge this in and then tack it on.

cdubz:
@jcgoette is this ready for review again? I forgot about it (:

jcgoette:
I'm over in PR #189, but still working on referencing user model in dashboard cards base template. If you notice a simple problem, let me know. Else, I'll figure it out _eventually_. :)

Probably want to close this PR (#137).

cdubz:
Ah yes _that_ was my confusion! Thanks.

PR #138, opened 2020-03-26
Auto select "bottle" method for formula and fortified milk types

### LINKED ISSUE #127: Auto select "bottle" method with "formula" type
When a user selects "Formula" type for a breast feeding entry, the "bottle" method should be selected automatically. Some simple JS should be able to do this easily, but the process for pre-populating the last feeding type should also account for it.

phardy:
Pre-populates the method select on the feeding form, and adds a function to the form to change the method select if formula or fortified milk are selected. I've made a couple of assumptions here, please correct me if they're wrong and I'll happily update the PR.

- Both "formula" and "fortified milk" are expected to only be fed by bottle.
- If breast milk is selected, it's best _not_ to pre-populate the method select. That just... feels like better UX to me.

Closes #127 .

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -50,7 +50,7 @@ def set_initial_values(kwargs, form_type):
         if last_feeding:
             last_type = last_feeding.type
             last_feed_args = {'type': last_feeding.type}
-            if last_type == 'formula':
+            if last_type in ['formula', 'fortified breast milk']:
                 last_feed_args['method'] = 'bottle'
             kwargs['initial'].update(last_feed_args)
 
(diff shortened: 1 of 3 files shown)


cdubz:
Thanks for contributing! Tests will also need to be updated to account for (and verify) this behavior.

I need to look in to why tests don't run on PRs from other repos...

phardy:
Of course! I've fleshed out the feeding form test, which should cover cases where the method should and should not have an initial value.

But while I can write some basic javascript, I'm not at all familiar with how to test it. Some pointers to where (or if?) I should be looking for js tests here would be useful.

cdubz:
Right now there is no framework in place for testing JS in the application, so don't worry about that. Thanks for the initial values test, I'll take a look at the PR later this weekend.

cdubz:
Thanks, @phardy! This works great.

PR #143, opened 2020-05-17
Periodic session expiry update to prevent timeouts

phardy:
I did some experimentation using the `SESSION_SAVE_EVERY_REQUEST` setting, which does indeed end up updating the expiry on the sessionid cookie with every request. But quickly realised that saving the session on every request also means adding a write to the database to every request. I suppose it doesn't really matter for a small app like this, but that just didn't seem very efficient, so I thought I'd try to save it less frequently.

This PR adds new middleware that checks when the session was last updated, and resets the expiry at configurable periods, with a default of 24 hours. It does this by setting a timestamp in the user's session.

### DIFF
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -23,6 +23,7 @@ class UserTimezoneMiddleware:
                 pass
         return self.get_response(request)
 
+
 class RollingSessionMiddleware:
     """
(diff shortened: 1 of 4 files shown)


cdubz:
Thanks, @phardy! Changes look good.

PR #144, opened 2020-05-18
Ensure date passed to nap filter is localtime.

### LINKED ISSUE #136: "Today's Naps" card no longer has data
Looks like a recent regression, but haven't had a chance to look in to it more.

phardy:
The nap card was taking a UTC date and passing it to the filter without TZ information, so it was being interpreted as a localtime date. This PR changes the card to use the same date lookup method as other cards that need it (Today's Sleep, Today's Tummy Time).

But this is a code path that a few dashboard cards use, and it isn't really being tested. I'm not sure how to properly test it either, short of mocking `timezone.localtime()`.

Closes #136

### DIFF
--- a/dashboard/templatetags/cards.py
+++ b/dashboard/templatetags/cards.py
@@ -148,8 +148,9 @@ def card_sleep_naps_day(child, date=None):
     :param date: a Date object for the day to filter.
     :returns: a dictionary of nap data statistics.
     """
-    date = timezone.localtime(date).astimezone(timezone.utc)
-    instances = models.Sleep.naps.filter(child=child, start__date=date.date())
+    if not date:
+        date = timezone.localtime().date()
+    instances = models.Sleep.naps.filter(child=child, start__date=date)
     return {
         'type': 'sleep',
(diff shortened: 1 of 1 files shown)


cdubz:
Thanks for working on this, @phardy! My partner has been especially annoyed by this over the past few months but I haven't had time to dig in to it, hah. I will review this tonight and see if I can figure out how to test.

In the past I had used [`override_settings`](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#overriding-settings) to do some TZ testing, but I pulled that out when I introduced the User based TZ setting.

cdubz:
Looks good! Thanks again, @phardy. I don't think it's worth getting in to the tests right now because they aren't very effective in the first place. Another "some day" task...

phardy:
Actually, if we remove the parameter check and just go with something similar to what the original code was doing like:

    date = timezone.localtime().date(date)

that'll get rid of the untested code path. After that, I think just testing the function with a couple of different TZ-aware datetime objects as parameters would be enough?

phardy:
Heh, fair enough!

Issue #148, opened 2020-07-08
Using 24H format for time


cdubz:
Hi! I’m having the reverse problem: I would like to not use AM/PM but 24H format. I’m using the app in English.

My first thought was `DATETIME_INPUT_FORMATS`, but that is about the input format, not display format. I then thought of setting [DATE_FORMAT](https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-DATE_FORMAT) but it seems that it will get overridden anyhow because `USE_L10N` is enabled? (that, and I would presumably need to set `DATETIME_FORMAT`, `TIME_FORMAT` and `SHORT_DATE_FORMAT`)?

Digging further, it appears that `USE_L10N` is conditioned by `LANGUAGE_CODE` ; however that is hardcoded to 'en' (presumably aliased somewhere by Django to `en-us`?) in `babybuddy/babybuddy/settings/base.py`  so I guess it would not be taken into account either?).

Am I missing something?

_Originally posted by @JeanFred in https://github.com/babybuddy/babybuddy/issues/93#issuecomment-654884934_

cdubz:
Will test this when I get a chance. There is also [`babybuddy.formats`](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/formats/en/formats.py) for _input_ formats, which seems like it may also be relevant. In theory this should definitely be something Baby Buddy supports, but right now the setup may be totally reliant on L10n formatting. And with the frontend date picker stuff, this may need to be a feature rather than something that can be just customized in settings.

JeanFred:
Thanks for splitting this into its own issue and having looked into it!

cdubz:
I also prefer 24 hour time formats, so I am interested in supporting this (: It's a bit more complicated that Django settings can handle because of the locale support, but doable as a global app setting (per-user setting didn't have a clear path because of the way Django's locale-based time formats work).

I got this as far along as I had time for in #150. In some basic tests it appears to be functional, but documentation and _actual_ tests are still needed.

PR #150, opened 2020-07-22
Add support for 24 hour time override (#148)

cdubz:
See #148. Still needs:

- [x] Support for [`SHORT_DATE_FORMAT`](https://docs.djangoproject.com/en/3.0/ref/settings/#short-datetime-format). _I missed this initially and don’t think it is used anywhere but it should be handled for completeness._
- [x] Documentation on the new setting
- [x] Tests

A number of tests currently fail when `USE_24_HOUR_TIME_FORMAT=True`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ take precedence over the contents of an `.env` file.**
 - [`NAP_START_MIN`](#nap_start_min)
 - [`SECRET_KEY`](#secret_key)
 - [`TIME_ZONE`](#time_zone)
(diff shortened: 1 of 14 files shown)


cdubz:
A test has been added for the new `USE_24_HOUR_TIME_FORMAT` setting, however it can only be run in isolation because of how it affects Django internals with respect to locales. A solution to this would probably involve refactoring the gulp testing commands quite a bit, so I may just merge this in as-is and not worry about the missed test for now.

PR #155, opened 2020-08-07
Updates babybuddy to use MySQL as a backend database.

### LINKED ISSUE #154: How to configure for MySQL?
I see that the original install used postgres, and there was a new docker-compose file supplied for sqlite. In the readme it mentions you can use MySQL. How do I go about configuring this in the docker-compose?

Thanks!

sstratoti:
Resolves #154

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,12 @@
 FROM python:3 as app
 ENV PYTHONUNBUFFERED 1
 RUN pip install --upgrade pipenv gunicorn
+RUN pip install pymysql
 WORKDIR /app
 COPY Pipfile /app/
 COPY Pipfile.lock /app/
 RUN pipenv install --deploy --system
+RUN pipenv install mysqlclient
 ADD manage.py /app/
 ADD api /app/api
 ADD babybuddy /app/babybuddy
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,6 @@ WORKDIR /app
 COPY Pipfile /app/
 COPY Pipfile.lock /app/
 RUN pipenv install --deploy --system
-RUN pipenv install mysqlclient
 ADD manage.py /app/
 ADD api /app/api
 ADD babybuddy /app/babybuddy
--- a/babybuddy/__init__.py
+++ b/babybuddy/__init__.py
@@ -52,7 +52,3 @@ __license__ = 'BSD 2-Clause'
 VERSION = __version__
 
 default_app_config = 'babybuddy.apps.BabyBuddyConfig'
-
-import pymysql
-
-pymysql.install_as_MySQLdb()
--- a/babybuddy/__init__.py
+++ b/babybuddy/__init__.py
@@ -52,3 +52,7 @@ __license__ = 'BSD 2-Clause'
 VERSION = __version__
 
 default_app_config = 'babybuddy.apps.BabyBuddyConfig'
+
+import pymysql
+
+pymysql.install_as_MySQLdb()
--- a/docker-compose.example.mysql.yml
+++ b/docker-compose.example.mysql.yml
@@ -12,7 +12,7 @@ services:
     volumes:
       - data:/app/data:rw
       - media:/app/media:rw
-      - /path/to/my.cnf:/app/my.cnf
+      - /path/to/my.cnf:/app/data/my.cnf
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
       - "8000:8000"
--- a/docker-compose.example.mysql.yml
+++ b/docker-compose.example.mysql.yml
@@ -1,0 +1,21 @@
+version: "3.7"
+services:
+  app:
+    image: babybuddy/babybuddy
+    # See README.md#configuration for other environment configuration options.
+    environment:
+      - ALLOWED_HOSTS=
+      - DJANGO_SETTINGS_MODULE=babybuddy.settings.mysql
+      - SECRET_KEY=
+      - TIME_ZONE=
+      - DEBUG=False # Turn to False in production
+    volumes:
+      - data:/app/data:rw
+      - media:/app/media:rw
+      - /path/to/my.cnf:/app/my.cnf
+    command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
+    ports:
+      - "8000:8000"
+volumes:
+  data: {}
+  media: {}
--- a/docker-compose.example.mysql.yml
+++ b/docker-compose.example.mysql.yml
@@ -12,7 +12,6 @@ services:
     volumes:
       - data:/app/data:rw
       - media:/app/media:rw
-      - /path/to/my.cnf:/app/data/my.cnf
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
       - "8000:8000"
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -1,10 +1,12 @@
 FROM python:3 as app
 ENV PYTHONUNBUFFERED 1
 RUN pip install --upgrade pipenv gunicorn
+RUN pip install pymysql
 WORKDIR /app
 COPY Pipfile /app/
 COPY Pipfile.lock /app/
 RUN pipenv install --deploy --system
+RUN pipenv install mysqlclient
```
cdubz, line 9: @sstratoti `pymysql` replaces this requirement, so we don't need to run this here.

Dockerfile
```diff
@@ -1,10 +1,12 @@
 FROM python:3 as app
 ENV PYTHONUNBUFFERED 1
 RUN pip install --upgrade pipenv gunicorn
+RUN pip install pymysql
```
cdubz, line 4: The PostgreSQL library is installed as part of the base Pipenv requirements and I _think_ we can do the same thing with the `pymysql` library because, unlike `mysqlclient`, it doesn't need to compile against mysql client libraries.

So lets get this instead in to Pipenv, which you can do by running `pipenv install pymysql` from the repo root and committed the changes to `Pipfile` and `Pipfile.lock`.

babybuddy/__init__.py
```diff
@@ -52,3 +52,7 @@
 VERSION = __version__
 
 default_app_config = 'babybuddy.apps.BabyBuddyConfig'
+
+import pymysql
+
+pymysql.install_as_MySQLdb()
```
cdubz, line 58: This would run for all configurations of the the app, regardless of the database used and that is not ideal. I _think_ it would be safe to move this `babybuddy/settings/mysql.py` but I'm not totally sure. I'm unclear on what the `install_as_MySQLdb` function does and if it would be OK to run multiple times?

sample-my.cnf
```diff
@@ -0,0 +1,8 @@
+[client]
+database = babybuddy
+host = localhost
+user = babybuddy
+password = babybuddy
+default-character-set = utf8
+use_unicode = True
+init_command = 'SET default_storage_engine=INNODB'
```
cdubz, line 8: Lets move this file to the `etc` dir in the repo root. This is where configuration stuff is stored.

docker-compose.example.mysql.yml
```diff
@@ -0,0 +1,21 @@
...
+    environment:
+      - ALLOWED_HOSTS=
+      - DJANGO_SETTINGS_MODULE=babybuddy.settings.mysql
+      - SECRET_KEY=
+      - TIME_ZONE=
+      - DEBUG=False # Turn to False in production
+    volumes:
+      - data:/app/data:rw
+      - media:/app/media:rw
+      - /path/to/my.cnf:/app/data/my.cnf
```
cdubz, line 15: Can this example just be `- data/my.cnf:/app/data/my.cnf` or does it need to be this way? I was kind of hoping we could guide people to just drop `my.cnf` _in to_ the `data` directory so it would be picked up there without an extra `volumes` line here.

Dockerfile
sstratoti: So when I try to run this, I get the following output:
```
Creating a virtualenv for this project…
Using /usr/bin/python3 (3.8.2) to create virtualenv…
⠋/usr/bin/python3: No module named pipenv.pew

Virtualenv location: 
Installing pymysql…
⠼
Error:  An error occurred while installing pymysql!
/bin/sh: 1: /bin/pip: not found
```

sstratoti: I went back to the "Installation" portion of the contributing doc, and found the steps to setup the environment.

I'm trying to run 
```
# pipenv install --three
Creating a virtualenv for this project…
Using /usr/bin/python3 (3.8.2) to create virtualenv…
⠋/usr/bin/python3: No module named pipenv.pew

Virtualenv location: 
Creating a virtualenv for this project…
Using /usr/bin/python3 (3.8.2) to create virtualenv…
⠙/usr/bin/python3: No module named pipenv.pew

Virtualenv location: 
Installing dependencies from Pipfile.lock (f8bb4c)…
Traceback (most recent call last):▉▉▉▉▉ 0/42 — 00:00:00
  File "/usr/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==11.9.0', 'console_scripts', 'pipenv')()
  File "/usr/lib/python3/dist-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3/dist-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pipenv/cli.py", line 349, in install
    core.do_install(
  File "/usr/lib/python3/dist-packages/pipenv/core.py", line 1875, in do_install
    do_init(
  File "/usr/lib/python3/dist-packages/pipenv/core.py", line 1362, in do_init
    do_install_dependencies(
  File "/usr/lib/python3/dist-packages/pipenv/core.py", line 832, in do_install_dependencies
    c = pip_install(
  File "/usr/lib/python3/dist-packages/pipenv/core.py", line 1477, in pip_install
    c = delegator.run(pip_command, block=block)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/delegator.py", line 267, in run
    c.run(block=block, binary=binary)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/delegator.py", line 156, in run
    s = PopenSpawn(self._popen_args, **pexpect_kwargs)
  File "/usr/lib/python3/dist-packages/pipenv/vendor/pexpect/popen_spawn.py", line 46, in __init__
    self.proc = subprocess.Popen(cmd, **kwargs)
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/bin/pip'
     ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/42 — 00:0
```

babybuddy/__init__.py
sstratoti: This worked! I'll commit it when I figure out the pipenv error above

Dockerfile
cdubz: Looks like pip is missing? What do you get for the following commands? --

```
python -v
pip -v
python3 -v
pip3 -v
pipenv -v
```

sstratoti: **python -V**
Python 2.7.18rc1
**pip -v** 
not found
**python3 -V**
Python 3.8.2
**pip3 -V**
pip 20.0.2 from /usr/lib/python3/dist-packages/pip (python 3.8)
**pipenv --version**
pipenv, version 11.9.0

cdubz: Hm, I have seen people in the past have to alias `python` to `python3` to make Pipenv happy. You could try ` alias python=python3` before running the commands (and maybe also `alias pip=pip3`).

Alternatively maybe take a look at this reference for the `No module named pipenv.pew` error -- which I have never seen before: pypa/pipenv/issues/2557

cdubz:
Closing this as outdated. There is more recent discussion in #154.

PR #171, opened 2020-10-24
Refactor tests_format

### LINKED ISSUE #170: Use of datetime.date.fromisoformat breaks Python 3.6 compatability
This is only used in [babybuddy/tests/tests_formats.py#L51](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/tests/tests_formats.py#L51) so refactoring it away hopefully won't be too difficult.

This was discovered when #158 was resolved.

killua1zoldyck:
Fix #170

### DIFF
--- a/babybuddy/tests/tests_formats.py
+++ b/babybuddy/tests/tests_formats.py
@@ -47,14 +47,3 @@ class FormatsTestCase(TestCase):
 
         with self.assertRaises(ValidationError):
             field.to_python('invalid date string!')
-
-        dt = datetime.datetime.fromisoformat('2011-11-04 23:05:59')
-        self.assertEqual(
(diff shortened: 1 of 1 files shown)


cdubz (changes requested):
@killua1zoldyck, thanks for grabbing this -- but we do actually want to keep the tested functionality here. We just need to change how we create the `datetime` objects on Lines 51 and 55 to not use `fromisoformat`. E.g. Line 51 could be changed to `dt = datetime.datetime(2011, 11, 4, 23, 5, 59)` and Line 55 can be changed similarly for its date.

cdubz:
I made this change with a direct commit as it was interfering with other branches. Thanks though!

Issue #165, opened 2020-09-12
Adding via Portainer


Cpainy:
Is it possible to add babybuddy via portainer? I use this to manage all my containers.
I'm getting errors trying to deploy babybuddy with portainer. 
I can't change the docker-compose.example.yml with portainer I think?

Regards

cdubz:
I’ve never heard of or used Portainer, but I can take a look at some point. Can you throw up some links for relevant documentation?

Cpainy:
It's one of many docker container management systems. Lets you install containers via a GUI. The setup process is pretty similar between different system.
Much easier for noobs like me. This is the first container I had trouble setting up because I couldn't get it to run via Portainer.

Eventually I did manage to install Babybuddy via docker using te CLI for my ubuntu server but I can't figure out how to set up an IP and change the ports so babybuddy is accessible to other PC's on the network. 
After that Portainer detected the baby buddy app and database and in the logs I can see the the local 0.0.0.0:8000 ip but I can access it from another PC.

Documentation here: [https://www.portainer.io/documentation/](https://www.portainer.io/documentation/)

ssilverr:
I tried deploying it as a stack using version: "2" as that's the docker-compose version that Portainer supports. Basically just copy the , but I get an error:
`standard_init_linux.go:211: exec user process caused "exec format error"`

Do note, I'm noob, I just started my journey to selfhosted-dom.

cdubz:
@ssilverr I'm not sure what it is exactly, but I believe something in the Docker Compose file is version 3+ only.

ssilverr:
@cdubz bummer. I'm not sure how to get it up runnig with Portainer and Pi. @Cpainy did you manage to do it?

Cpainy:
I did get it to work in Portainer, but only after installing it first via command line interface and docker compose as instructed in the documentation of babybuddy.
Portainer just picked it up automatically after that and I can use it to change settings now.

phidauex:
For future users, below is an example of a Compose file that works in Portainer 2.0. It is just a slight modification from the SQLite compose file.

Volumes may need editing, and you would need to add a network if you want it to join a particular subnet.

Note the change to port 7000 - Portainer uses 8000 for its own web UI.

`version: 2
services:
  app:
    image: babybuddy/babybuddy
    container_name: "babybuddy"
    # See README.md#configuration for other environment configuration options.
    environment:
      - ALLOWED_HOSTS=localhost
[...]

PR #179, opened 2020-12-30
Updates to docker compose files for Portainer compatibility

phidauex:
Hi, this pull request is not 100% ready - I'm still a github novice so I'm probably doing something wrong with my commits.

The changes to `docker-compose.example.sqlite.yml` are good, and tested in Portainer. I dropped to the earliest compatible compose version (2.4), added some notes, and a friendly container name. This addresses issue: https://github.com/babybuddy/babybuddy/issues/165

The changes to `docker-compose.example.yml` are NOT working in Portainer, and I'm not sure why. The startup script appears to be looking for a service at hostname "db", but the hostname will either be generated from the stack "babybuddy_db_1" or from the container_name "babybuddydb".

I'm not sure how to separate the commits (again, quite novice at this part), but at least the first part appears to be good, and shouldn't cause any breaking changes with standard compose deployments.

### DIFF
--- a/docker-compose.example.sqlite.yml
+++ b/docker-compose.example.sqlite.yml
@@ -20,6 +20,7 @@ services:
       - "8000:8000"
     # For Portainer deployment, select an alternate port for babybuddy, such as port 7000
     # - "7000:8000"
+    restart: unless-stopped
 volumes:
   data: {}
   media: {}
--- a/docker-compose.example.sqlite.yml
+++ b/docker-compose.example.sqlite.yml
@@ -4,7 +4,7 @@ services:
     image: babybuddy/babybuddy
     # Container_name is an optional value for single-container deployments to set a friendly container name.
     # Disable in swarm or other multi-container setups.
-    container_name: babybuddy
+    container_name: babybuddyapp
     # See README.md#configuration for other environment configuration options.
     environment:
       - ALLOWED_HOSTS=localhost   # comma separated list of IP addresses or hosts that can access the web UI
@@ -17,9 +17,7 @@ services:
       - media:/app/media:rw
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
-      - "8000:8000"
-    # For Portainer deployment, select an alternate port for babybuddy, such as port 7000
-    # - "7000:8000"
+      - "8000:8000" # For Portainer, select an alternate port for babybuddy, such as 7000, for example "7000:8000"
     restart: unless-stopped
 volumes:
   data: {}
--- a/docker-compose.example.sqlite.yml
+++ b/docker-compose.example.sqlite.yml
@@ -2,9 +2,9 @@ version: "2.4"
 services:
   app:
     image: babybuddy/babybuddy
-    # Container_name is an optional value for single-container deployments to set a friendly container name.
-    # Disable in swarm or other multi-container setups.
-    container_name: babybuddyapp
+    container_name: babybuddyapp   # Container_name is an optional value for single-container
+                                   # deployments to set a friendly container name.
+                                   # Disable in swarm or other multi-container setups.
     # See README.md#configuration for other environment configuration options.
     environment:
       - ALLOWED_HOSTS=localhost   # comma separated list of IP addresses or hosts that can access the web UI
@@ -17,7 +17,7 @@ services:
       - media:/app/media:rw
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
-      - "8000:8000" # For Portainer, select an alternate port for babybuddy, such as 7000, for example "7000:8000"
+      - "8000:8000"               # For Portainer, select another port for babybuddy such as 7000, for example "7000:8000"
     restart: unless-stopped
 volumes:
   data: {}
--- a/docker-compose.example.sqlite.yml
+++ b/docker-compose.example.sqlite.yml
@@ -1,20 +1,25 @@
-version: "3.7"
+version: "2.4"
 services:
   app:
     image: babybuddy/babybuddy
+    # Container_name is an optional value for single-container deployments to set a friendly container name.
+    # Disable in swarm or other multi-container setups.
+    container_name: babybuddy
     # See README.md#configuration for other environment configuration options.
     environment:
-      - ALLOWED_HOSTS=
+      - ALLOWED_HOSTS=localhost   # comma separated list of IP addresses or hosts that can access the web UI
       - DJANGO_SETTINGS_MODULE=babybuddy.settings.base
-      - SECRET_KEY=
-      - TIME_ZONE=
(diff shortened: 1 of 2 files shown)


docker-compose.example.sqlite.yml
```diff
@@ -1,20 +1,26 @@
-version: "3.7"
+version: "2.4"
```
cdubz, line 1: Any particular reason for 2.4 vs. some other version? Is that the minimum for the features use by the compose file?

docker-compose.example.sqlite.yml
```diff
@@ -1,20 +1,26 @@
...
 services:
   app:
     image: babybuddy/babybuddy
+    # Container_name is an optional value for single-container deployments to set a friendly container name.
+    # Disable in swarm or other multi-container setups.
+    container_name: babybuddy
     # See README.md#configuration for other environment configuration options.
     environment:
-      - ALLOWED_HOSTS=
+      - ALLOWED_HOSTS=localhost   # comma separated list of IP addresses or hosts that can access the web UI
```
cdubz, line 10: Does `localhost` make this "just work" with compose? If so this seems like a good change.

docker-compose.example.sqlite.yml
```diff
@@ -1,20 +1,26 @@
...
+      - DEBUG=False               # Turn to False in production
     volumes:
       - data:/app/data:rw
       - media:/app/media:rw
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
       - "8000:8000"
+    # For Portainer deployment, select an alternate port for babybuddy, such as port 7000
+    # - "7000:8000"
+    restart: unless-stopped
```
cdubz, line 23: What does this setting do?

docker-compose.example.yml
```diff
@@ -1,28 +1,38 @@
...
       - PGDATA=/db-data
       - POSTGRES_PASSWORD=postgres # has to correspond with POSTGRES_PASSWORD in APP
     volumes:
       - db:/db-data:rw
+    restart: unless-stopped
   app:
     image: babybuddy/babybuddy
+    # Container_name is an optional value for single-container deployments to set a friendly container name.
+    # Disable in swarm or other multi-container setups.
+    container_name: babybuddy
```
cdubz, line 18: In keeping the style of the `db` service, lets name this `babybuddyapp`.

docker-compose.example.sqlite.yml
```diff
@@ -1,20 +1,26 @@
...
+      - SECRET_KEY=               # Generate a random string here to secure the Django instance
+      - TIME_ZONE=                # In the tzdata format, IE, "America/Denver"
+      - DEBUG=False               # Turn to False in production
     volumes:
       - data:/app/data:rw
       - media:/app/media:rw
     command: bash -c 'python manage.py migrate --noinput && python manage.py createcachetable && gunicorn babybuddy.wsgi -b :8000 --log-level=info'
     ports:
       - "8000:8000"
+    # For Portainer deployment, select an alternate port for babybuddy, such as port 7000
```
cdubz, line 21: This would also require commenting out the port above, right? I think I'd rather see this information in the README instead of as a comment here. It could be a small additional section under the [Docker section](https://github.com/babybuddy/babybuddy/blob/master/README.md#docker).

cdubz (changes requested):
Thanks for taking the time to learn about this and contribute! I have made a few change requests/asked a few questions in line comments. Let me know if you need any guidance making further changes.

docker-compose.example.sqlite.yml
phidauex: 2.4 is the latest in the 2.x branch, and fully supports all the options, as well as being Portainer (libcompose) compatible. Once Portainer's main branches are all updated this can probably come up to the 3.x branch, but honestly most of the new compose features in 3.x are heavily oriented toward load sharing/cpu/multicontainer options that probably aren't very important for bb users.

docker-compose.example.sqlite.yml
phidauex: I haven't tested all the permutations of this, but at least for someone running it on their own machine "localhost" would "just work". Most other users will have to add an additional comma separated host. The debug mode in Django seems to allow for [*], but that seemed like an unwise default for production.

docker-compose.example.sqlite.yml
phidauex: `restart:` sets how the container should behave if it is stopped by another process, or if the container crashes. `unless-stopped` means the container will be automatically restarted if it stops for any reason, other than if the user stopped it intentionally. I've found it a common addition to compose files for services that you would want to restart if they crashed or the machine reboots. https://docs.docker.com/compose/compose-file/compose-file-v2/#restart

docker-compose.example.yml
phidauex: Agreed, updated!

docker-compose.example.sqlite.yml
phidauex: See updated alternate - instead of a commented option, I just reference the note in an inline comment. Does that look OK or should we leave it in the docs only?

docker-compose.example.yml
```diff
@@ -1,28 +1,36 @@
...
+    restart: unless-stopped
   app:
     image: babybuddy/babybuddy
+    # Container_name is an optional value for single-container deployments to set a friendly container name.
+    # Disable in swarm or other multi-container setups.
+    container_name: babybuddyapp
     # See README.md#configuration for other environment configuration options.
     environment:
-      - ALLOWED_HOSTS=
-      - DB_PASSWORD=postgres # has to correspond with POSTGRES_PASSWORD in DB
```
cdubz, line 15: One small thing -- this is a recent change that should be kept, please revert this line and remove the old `POSTGRES_PASSWORD` below (L25). And the new inline comment should read, "Must correspond with POSTGRES_PASSWORD in DB"

docker-compose.example.yml
```diff
@@ -1,28 +1,36 @@
...
+version: "2.4"
 services:
   db:
     image: postgres:11 # pin postgres to a major version
+    # Container_name is an optional value for single-container deployments to set a friendly container name.
+    # Disable in swarm or other multi-container setups.
+    container_name: babybuddydb
     environment:
       - PGDATA=/db-data
       - POSTGRES_PASSWORD=postgres # has to correspond with DB_PASSWORD in APP
```
cdubz, line 10: While we're at it, lets update this inline comment to match the spacing and format of the others (i.e. "Must correspond with DB_PASSWORD in APP"

cdubz (changes requested):
Thanks for the updates! I've requested two more small changes related to the rebase on recent commits to `master`.

docker-compose.example.yml
phidauex: I see, corrected, thanks!

docker-compose.example.yml
phidauex: See updated formatting, trying to make things "clean". I like inline documentation but it can get a bit busy...

cdubz (approved):

cdubz:
Looks great. Thanks for the help!

Issue #180, opened 2020-12-31
Postgres password concern


Zutart:
My deployment strategy has multiple docker containers utilizing a single postgres docker container for their database needs. I've created individual users for each service and attached them to their respective tables to ensure no one service could have access to another.

It appears as if this container is hard coded to utilize the root password for the postgres container. Am I incorrect in that assessment? If not, would there be a way to pass both a username and password (and potentially the rest of the variables, such as database name and port)?

cdubz:
Do you mean the settings in the [docker settings file](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/settings/docker.py)? That file looks for a `POSTGRES_PASSWORD` environment variable for the password. Currently, setting the user by environment variable is not supported, but that would be easy enough to change. Would that work for your use case?

Zutart:
I think it would. Essentially, this is an example of another docker image and how i configure its access to postgres:

      - DB_TYPE=postgres
      - DB_HOST=database.postgres:5432
      - DB_NAME=gitea
      - DB_USER=gitea
      - DB_PASSWORD={password}

That way I can specify a specific server, port, username, pw, and db if any aren't the default values expected and I can better control access to the DB without needing to expose the root password. It may also help with the foundation for this image to be able to use other DB types like MySQL.

Let me know if that helps to better explain my thinking, and appreciate you responding so quickly!

cdubz:
Yeah, I like that approach (exposing all the DB settings as ENV variables). This will require documentation updates as well and we'll want to continue to support `POSTGRES_PASSWORD` until a new major version release. Are you interested in making a PR? No worries if not, I wouldn't mind learning a bit about how/if Python and Django handle deprecation (i.e. for the old env variable).

Zutart:
Actually I would love to, I've only recently been able to fully dive into this world and this is a project I'd love to contribute to given my second is on the way and I REALLY could have used it with number one.

I'll go ahead and get a branch moving on this shortly once I have a break at work and if I can get it figured out (or if someone else can help) we'll get this updated that way.

cdubz:
Cool, thanks. I'll be happy to help if you run in to any trouble. I think the three settings files we'd want to make changes to would be:

1. [base.py](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/settings/base.py)
2. [docker.py](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/settings/docker.py)
[...]

PR #182, opened 2021-01-03
Allow for custom PostgreSQL Deployments

Zutart:
Fixes issue #180 by allowing for any of the new PostgreSQL variables to be defined by user or fallback to expected defaults from existing deployment examples.

### DIFF
--- a/babybuddy/settings/docker.py
+++ b/babybuddy/settings/docker.py
@@ -10,7 +10,7 @@ DATABASES = {
         'ENGINE': os.getenv('DB_ENGINE') or 'django.db.backends.postgresql_psycopg2',
         'NAME': os.getenv('DB_NAME') or 'postgres',
         'USER': os.getenv('DB_USER') or 'postgres',
-        'PASSWORD': os.environ.get('DB_PASSWORD'),
+        'PASSWORD': os.environ.get('DB_PASSWORD') or os.environ.get('POSTGRES_PASSWORD'),
         'HOST': os.getenv('DB_HOST') or 'db',
         'PORT': os.getenv('DB_PORT') or 5432,
     }
--- a/babybuddy/settings/docker.py
+++ b/babybuddy/settings/docker.py
@@ -7,11 +7,11 @@ from .base import *
 # Load settings from env file / variables with fallback defaults to support current psql deployment
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': os.getenv('POSTGRES_DB') if os.getenv('POSTGRES_DB') else 'postgres',
-        'USER': os.getenv('POSTGRES_USER') if os.getenv('POSTGRES_USER') else 'postgres',
-        'PASSWORD': os.environ.get('POSTGRES_PASSWORD'),
-        'HOST': os.getenv('POSTGRES_HOST') if os.getenv('POSTGRES_HOST') else 'db',
-        'PORT': os.getenv('POSTGRES_PORT') if os.getenv('POSTGRES_PORT') else 5432,
+        'ENGINE': os.getenv('DB_ENGINE') or 'django.db.backends.postgresql_psycopg2',
+        'NAME': os.getenv('DB_NAME') or 'postgres',
+        'USER': os.getenv('DB_USER') or 'postgres',
+        'PASSWORD': os.environ.get('DB_PASSWORD'),
+        'HOST': os.getenv('DB_HOST') or 'db',
+        'PORT': os.getenv('DB_PORT') or 5432,
     }
 }
--- a/babybuddy/settings/docker.py
+++ b/babybuddy/settings/docker.py
@@ -4,13 +4,14 @@ from .base import *
 # Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
+# Load settings from env file / variables with fallback defaults to support current psql deployment
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'postgres',
-        'USER': 'postgres',
+        'NAME': os.getenv('POSTGRES_DB') if os.getenv('POSTGRES_DB') else 'postgres',
+        'USER': os.getenv('POSTGRES_USER') if os.getenv('POSTGRES_USER') else 'postgres',
         'PASSWORD': os.environ.get('POSTGRES_PASSWORD'),
-        'HOST': 'db',
-        'PORT': 5432,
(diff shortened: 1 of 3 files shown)


babybuddy/settings/docker.py
```diff
@@ -4,13 +4,14 @@
 # Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
+# Load settings from env file / variables with fallback defaults to support current psql deployment
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'postgres',
-        'USER': 'postgres',
+        'NAME': os.getenv('POSTGRES_DB') if os.getenv('POSTGRES_DB') else 'postgres',
```
cdubz, line 11: Lets use the `or` keyword here instead of inline `if-else` (also for L12,14-15).

E.g. `os.environ.get('POSTGRES_DB') or 'postgres'`

babybuddy/settings/docker.py
```diff
@@ -4,13 +4,14 @@
 # Database
 # https://docs.djangoproject.com/en/3.0/ref/settings/#databases
 
+# Load settings from env file / variables with fallback defaults to support current psql deployment
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
```
cdubz, line 10: We also want this to be configurable, so we should use e.g. `DB_ENGINE`.

And similarly, lets use the `DB_` prefix for all of these settings --

- `DB_ENGINE`
- `DB_NAME` (for `POSTGRES_DB`)
- `DB_USER`
- `DB_HOST`
- `DB_PORT`

cdubz (changes requested):
Thanks @Zutart for getting this started! I have made a few comments with change requests -- primary one is that I'd like to use a ` DB_` prefix instead of `POSTGRES_` for the environment variables so we could theoretically use them in other configuration situations.

babybuddy/settings/docker.py
Zutart: Makes a ton of sense and I'm making the edits now. One question I had was about your plan on handling the rollout for additional database types and how best to lay the foundation for that moving forward. I've seen examples of people checking the db_engine variable and wrapping the proper database structure within an if/else statement right inside of the settings.py file, and I've seen people break out the different DB structures into other files like settings.psql.py that would supercede the original settings. I'm assuming the latter would require the app to check the env and call the proper settings file somewhere else (or put that on the user to issue as a command when creating their docker container) whereas the former could theoretically handle the adjustments without having to make additional code changes.

Would you have a preference on how you want to proceed?

babybuddy/settings/docker.py
Zutart: Updated in 6f83e16. Also added in check for DB_ENGINE.

Zutart: Marking this unresolved as I just thought of another problem / question - are you going to put the exact string require by django for the db_engine type (i.e. django.db.backends.mysql) in the documentation for users to fill out or would there be an elegant way to check the variable and select the right engine string based on easier expected variables (i.e. mysql)?

cdubz: I think for a first pass, lets just have the let users fill in the `DB_ENGINE` string as desired. The Docker settings file should default to the currently value (`django.db.backends.postgresql_psycopg2`) and the documentation can be updated to reference https://docs.djangoproject.com/en/3.0/ref/settings/#engine for other standards.

Zutart: Updates pushed in 0c05cd3 with README.md changes reflected

babybuddy/settings/docker.py
Zutart: Changes pushed in 0c05cd3 and 6f83e16, discussion about above question took place in the other conversation

babybuddy/settings/docker.py
```diff
@@ -4,13 +4,14 @@
...
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'postgres',
-        'USER': 'postgres',
-        'PASSWORD': os.environ.get('POSTGRES_PASSWORD'),
-        'HOST': 'db',
-        'PORT': 5432,
+        'ENGINE': os.getenv('DB_ENGINE') or 'django.db.backends.postgresql_psycopg2',
+        'NAME': os.getenv('DB_NAME') or 'postgres',
+        'USER': os.getenv('DB_USER') or 'postgres',
+        'PASSWORD': os.environ.get('DB_PASSWORD'),
```
cdubz, line 13: Two final items

1. Lets fall back on `POSTGRES_PASSWORD` for now. This way we don't break any existing docker compose configurations unexpectedly.
2. For new deployments, lets also update [`docker-compose.example.yml`](https://github.com/babybuddy/babybuddy/blob/master/docker-compose.example.yml) to use the new environment variable in the `app` section.

cdubz (changes requested):
@Zutart this looks great! Just two small changes requested then we can merge it in.

babybuddy/settings/docker.py
Zutart: Newest set of commits reflect those two changes

docker-compose.example.yml
```diff
@@ -4,18 +4,23 @@ services:
...
-      - POSTGRES_PASSWORD=postgres # has to correspond with POSTGRES_PASSWORD in APP
+      - POSTGRES_PASSWORD=postgres # has to correspond with DB_PASSWORD in APP
     volumes:
       - db:/db-data:rw
   app:
     image: babybuddy/babybuddy
     # See README.md#configuration for other environment configuration options.
     environment:
       - ALLOWED_HOSTS=
+      - DB_ENGINE=django.db.backends.postgresql
```
cdubz, line 15: Note that the default is actually `django.db.backends.postgresql_psycopg2`, not `django.db.backends.postgresql`. For the example file, lets stick to just providing the preset `DB_PASSWORD` variable.

cdubz (changes requested):
One more change needed...

docker-compose.example.yml
Zutart: Whoops, thanks for catching that. Updating now. Also assuming you meant to remove the default variables as examples and leave it as 

DB_HOST=

but if not just let me know and I'll update.

cdubz: No, I mean lets just replace the existing parameter there and not add the others -- the current default works correctly without any additional setup and anyone who wants further configurability can refer to the docs.

Zutart: Ahhhh ok, now I'm following. Latest commit should roll back those changes, but let me know if I'm still missing the mark.

cdubz (approved):

cdubz:
Looks good. Thanks for your contribution!

PR #188, opened 2021-01-18
Boolean field labels are no longer clickable

ntrecina:
### Summary
The label for boolean fields such as wet/solid on diaper changes is blank but still clickable which can result in accidentally toggling the value. This happens pretty commonly for me in the mobile view when closing the date time picker. I added a new class to boolean fields and applied css that disables pointer events so clicks have no effect.

#### Before
broken

#### After
fixed


cdubz:
Thanks for the detailed information and PR, @ntrecina! This is a nice fix.

PR #189, opened 2021-01-20
add option for hiding empty dashboard cards

jcgoette:
Picked up on work done by @BenjaminHae in PR #137.


jcgoette:
@cdubz, I'm not super familiar with Django, so thinking there must be a better way to do this instead of "getattr(Settings.objects.first(), 'dashboard_hide_empty')"? I'm open to suggestions.

dashboard/templatetags/cards.py
```diff
@@ -22,6 +22,9 @@ def card_diaperchange_last(child):
     """
     instance = models.DiaperChange.objects.filter(
         child=child).order_by('-time').first()
+
+    if not instance and getattr(Settings.objects.first(), 'dashboard_hide_empty'):
```
jcgoette, line 26: @cdubz, I'm not super familiar with Django, but I imagine there must be a better way to do this?

cdubz:
Hi, @jcgoette 👋 -- thanks for the PR! I haven't had a chance to review yet but should be able to do so this weekend. Note that this branch is failing CI due to linting issues. Take a look at e.g. [the Python 3.6 build](https://travis-ci.com/github/babybuddy/babybuddy/jobs/473439067) for details and see [CONTRIBUTING.md#lint](https://github.com/babybuddy/babybuddy/blob/master/CONTRIBUTING.md#lint) for details on how you can resolve those issue.

jcgoette:
Perfect, thanks.

Yes, wasn't sure how strict you were on line char limits. After you review
it may not be an issue anymore, but if it is I will fix. 👍🏻

dashboard/templatetags/cards.py
```diff
@@ -22,6 +22,9 @@ def card_diaperchange_last(child):
...
+        return {'hidden': True}
```
cdubz, line 27: @jcgoette I think you missed a step here. You are assigning this variable throughout but then not using it anywhere? You probably want to reference in to `dashboard/templates/cards/base.html` to decide whether or not to render the card.

dashboard/templates/cards/statistics.html
```diff
@@ -1,40 +1,46 @@
...
-            <div class="carousel-inner">
-                {% for stat in stats %}
-                    <div class="carousel-item{% if forloop.counter == 1 %} active{% endif %}">
-                        <span class="card-title">
-                            {% if stat.stat %}
-                                {% if stat.type == 'duration' %}
-                                    {{ stat.stat|duration_string:'m' }}
-                                {% elif stat.type == 'float' %}
-                                    {{ stat.stat|floatformat }}
+{% if stats|length > 0 and not user.settings.dashboard_hide_empty %}
```
cdubz, line 3: I don't think `stats|length` would ever be empty here? You'd still need to determine whether or not to hide this card in the same manner as the others. The basis would just be somewhat different in that you'd probably only want to hide it if none of the stats functions produce results. So slightly more complicated.

dashboard/templatetags/cards.py
cdubz: Yeah I don't love that pattern. My recommendation here would be to _not_ exit early in these functions with just the `hidden` prop but to add an `empty` prop to the regular returned object. Don't worry about `Settings.objects` at that point though, just use `user.settings.dashboard_hide_empty` when you check for the `empty` prop in `dashboard/templates/cards/base.html`.

cdubz (changes requested):
Thanks for contributing! This is pretty well along. See inline comments for requested changes. Please do also lint and ensure your code is passing linting tests.

I would also like to see tests for this behavior -- one in to `babybuddy/tests/tests_forms.py` to cover setting the open in user settings and at least one in `dashboard/tests/tests_templatetags.py` (in theory there should be one for each card for full coverage, but testing one example card would be a good start).

BenjaminHae:
I've picked up again what @jcgoette has done ;)
@cdubz when trying to access `user.settings.dashboard_hide_empt` in the base template, `user` is undefined. Do I need to import something somewhere? I'm not really familiar with Django.

cdubz:
@BenjaminHae Hmmm `user` should be defined. Are you sure it is not? See Django docs for reference: https://docs.djangoproject.com/en/3.2/topics/auth/default/#authentication-data-in-templates

BenjaminHae:
I've put the following code in `base.html`:
```
<div class="card-text">
    {% if user %}
       User exists
    {% else %}
            user does not exist
    {% endif %}
    {{ user.username }}
```
This is the result:
image

cdubz:
Hmmmm. I'll have to take a look at some point. You can see the `user` variable used, for example, in [babybuddy/templates/babybuddy/user_settings_form.html#L73](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/templates/babybuddy/user_settings_form.html#L73). I don't think anything special is done to support it's use there so it seems like it should be available elsewhere as well...

cdubz:
@BenjaminHae I see now -- the card templates and their base do not inherit from the system base templates. They are sort of "partial" templates because they are added via template tags in [dashboard/templates/dashboard/child.html](https://github.com/babybuddy/babybuddy/blob/master/dashboard/templates/dashboard/child.html) (as opposed to being rendered directly).

So the `user` context is available in the `child.html` template and you could work on this feature from there. If you do indeed need the `user` inside the partials you'll need to update the tag definitions in [dashboard/templatetags/cards.py](https://github.com/babybuddy/babybuddy/blob/master/dashboard/templatetags/cards.py) and pass in the necessary parameters when rendering those tags,

BenjaminHae:
@cdubz thank you very much, now I'm nearly finished, just need to find out how to mock the request context for the tests :)

cdubz:
Closing in favor of #213.

PR #201, opened 2021-04-04
Add solid food option to feeding types

0x4161726f6e:
I would find this helpful as I coordinate with day care


cdubz:
Thanks for the PR! I think this makes sense as a small and straightforward change to allow some new variation in the feeding types. This may have some effects on other areas of the application that expect the current options so we’ll need to evaluate that. But tests are passing (the Travis fail appears to be an unrelated issue with the Python 3.8 tests) so that’s a good sign (:

cdubz:
Thanks @0x4161726f6e! This is merged in to `master`. Will be in a release within the next few weeks or so.

Issue #205, opened 2021-04-25
Child slug may become empty


cben:
Currently it is easy for a child with non-ascii name to get very short or empty slug, as the actual characters get filtered out:
`'Мааян' 'Паскин-Чернявский'` => slug `'-'` :laughing: 
`'Мааян' 'Паскин'` => slug `''` :boom: 

In first case, it works but is silly and it's easy to get slug collisions.

Once any child has an empty slug, the app is "semi-bricked": many pages (like children list) break with NoReverseMatch errors:
> Reverse for 'child' with arguments '('',)' not found. 1 pattern(s) tried: ['children/(?P<slug>[-a-zA-Z0-9_]+)/$']

and you can't edit or even delete this child as she has no working URL...

- How to confirm this situation: visit `/api/children/` and look for empty slug.
- How to **un-brick** if you have admin account: visit `/admin/core/child/`, edit that child's name to include English letters.  The slug is always re-computed on save.

----

Proposed directions (sending PR for first 2 in a moment):

- [x] at least prevent setting such names
- [x] allow unicode chars in slugs
- [ ] A migration to fix at least empty slugs in existing installs?
- [ ] A migration to re-compute all slugs?  
  Not sure if a good idea: If they worked before, users may have already bookmarked per-child URLs that will now break.  New URLs will make more sense with non-ascii letters retained, yes, but who cares.  If people care, they can edit child name (even if no changes) and save to force re-computation.
- [ ] Let admin view show and/or manually edit slugs?
  - Manual editing is IMHO not worth it :-1:.  Raises question whether future `save()` by regular editing should preserve existing slug.  Currently `save()` always re-computes it from name, which would lose admin-set name.
  - Already shown in list, but not in edit form.  If not editable, showing read-only in edit form would be confusing as it wouldn't reflect that it's going to be re-computed; so showing only in list after save is perfect.

cdubz:
I don't think we need to put much effort in to a migration path here. Folks who hit this bug likely would have already either reached out for assistance or abandoned the app. I also don't think we need any level of admin edit for slugs.

I will leave this open for tests, though. I'd like to eventually add one to prove that the change in #206 was effective at supporting unicode slugs.

cben:
Just now got the dev env working.  Here is a start:
```diff
diff --git babybuddy/management/commands/fake.py babybuddy/management/commands/fake.py
index 5dac448..3173445 100644
--- babybuddy/management/commands/fake.py
[...]

PR #206, opened 2021-04-25
child slugs: Require non-blank, preserve unicode letters

cben:
Will prevent #205 from happening in future.  
`'Мааян' 'Паскин-Чернявский'` => slug `'мааян-паскин-чернявский'` :+1:  
This results in %-encoded URLs, but modern browsers display the unicode chars (e.g. https://babybuddy.../children/мааян-паскин-чернявский).

Does not fix existing installs, does not modify any existing slugs.  
slugs are currently re-computed on every save, so an existing child will get unicode chars included in slug on next edit (even if you don't change the name).  
If already broken, admin can still edit name.

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.template.defaultfilters import slugify
+from django.utils.text import slugify
 from django.utils import timezone
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
@@ -80,6 +80,7 @@ class Child(models.Model):
         verbose_name=_('Birth date')
     )
     slug = models.SlugField(
+        allow_unicode=True,
         blank=False,
         editable=False,
         max_length=100,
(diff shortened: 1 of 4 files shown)


cben:
temp app from this branch if you want to live test: https://babybuddy-unicode-slug-2dazlat.herokuapp.com/  (default user: admin/admin).
BTW, consider setting up a heroku pipeline with review apps (and preferrably DEBUG=True) to get these automatically for each PR here.  I did on my fork.

cben:
One alternative to %-encoded URLs would be approximate transliterating to ASCII.  Can do that if you prefer, https://stackoverflow.com/a/54578987/239657 has links to several libraries.

core/models.py
```diff
@@ -5,7 +5,7 @@
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.template.defaultfilters import slugify
+from django.utils.text import slugify
```
cben, line 8: confusingly, Django now has 2 `slugify()` functions, but only the utils.text one supports `allow_unicode` param.

cdubz:
@cben thanks for reporting this issue and submitting the PR — I like this approach of supporting unicode for the slugs (as opposed to ASCII conversion).

I’ll give a full review later today or tomorrow and think about what the migrate path will need to do. In the mean time we’ll also want tests to prove that these changes work. If you are able to add that to your PR that would help but I can also get them in after review.

Thanks, again.

cdubz:
Thanks again, @cben. This looks good. I'm not gonna sweat the merge path much here... will follow-up re: tests in #205.

jcgoette:
Did we miss the `makemigrations` for this PR?

cdubz:
We did. Thanks @jcgoette. Added in 71da84a.

Issue #202, opened 2021-04-08
Add LDAP as authentication backend


nitnelave:
Hello,

It would be great to have a possibility to re-use an existing LDAP backend for users. It shouldn't be too hard with django, hopefully.

Thanks,

BenjaminHae:
How about OpenID Connect?

cdubz:
There is a [django-auth-ldap/django-auth-ldap](https://github.com/django-auth-ldap/django-auth-ldap) plugin that would support this -- it looks well maintained but requires quite a bit of config. I don't have an LDAP server to test with but I'd be happy to consider a PR for this. Theoretically one could also just install the package on an existing BabyBuddy deployment and then configure it per their documentation.

nitnelave:
I could have a look, but I'm not familiar with Django. From the plugin's page, I need to add some configuration (in python), but I'm not sure where to add it. Babybuddy doesn't seem to have a settings.py or equivalent.

nitnelave:
Ah, I found it in babybuddy/settings/base.py. I'll have a look, and might also add authentication via REMOTE_USER header as well.

cdubz:
We have support for `REMOTE_USER` now via [`REVERSE_PROXY_AUTH`](https://docs.baby-buddy.net/configuration/security/#reverse_proxy_auth) so I think this use case is covered well enough.

PR #207, opened 2021-04-25
Add support for authentication based on LDAP or REMOTE_USER header

nitnelave:
Fixes https://github.com/babybuddy/babybuddy/issues/202

Disclaimer: This PR is mostly untested. I tried to run the usual `gulp lint` and `gulp test`, but I ran into issues setting up the environment, and this is more time than I want to put in this CL, since I'm not using the app anymore (sorry :/ ).

Anyway, here's my best effort, I hope it helps!

Note that it adds dependencies on python-ldap, which requires the `libsasl2-dev python-dev libldap2-dev libssl-dev` packages (debian/ubuntu) to be installed. The change should only affect devs, though, since the import is only attempted if the config option of "LDAP" is passed.

How to use:
If you set the env variable `AUTHENTICATION_METHOD=HEADER`, then django will expect the `REMOTE_USER` (and `REMOTE_GROUPS`) header, and identify based on that. If the headers are not present, it will fall back to normal authentication.

If you specify `AUTHENTICATION_METHOD=LDAP`, then you get the standard setup from https://github.com/django-auth-ldap/django-auth-ldap. The key points are:
  - `AUTH_LDAP_SERVER_URI`: the host
  - `AUTH_LDAP_BIND_DN`: the read user
  - `AUTH_LDAP_BIND_PASSWORD`: the password
  - `AUTH_LDAP_USER_DN_TEMPLATE`: where to find the users in LDAP
  - `AUTH_LDAP_GROUP_SCOPE`: the common root of the groups
  - (optional): `AUTH_LDAP_REQUIRE_GROUP`: the required group to be a babybuddy user
  - (optional): `AUTH_LDAP_USE_SPECIAL_GROUPS`: If this is true, this will look for the `active`, `staff`, `superuser` groups and map them to the django concepts.


nitnelave:
It was for my brother, but they ended up using another app they already had, because the mobile interface for baby buddy was not as good. I'm not sure what app, though.

cdubz:
@nitnelave If you happen to learn the name of the app they chose at some point I'd be interested to know. Baby Buddy serves a pretty niche audience of something like privacy-conscious users with a strong general interest in self-hosting so I don't really think of other similar apps as "competition" but I am always curious. Back when I initially developed this for our first child in 2017 we tried a lot of different apps and they were all seemingly terrible at the time, hah.

cdubz:
Closing this as outdated. Happy to consider it again if someone wants to take it up (:

PR #209, opened 2021-04-30
fix translation format strings

BenjaminHae:
fixes #208

PR #210, opened 2021-05-01
fixes translation (de)

BenjaminHae:
This PR fixes an error in the German translation introduced in #209 and contains the compiled translation.


cdubz:
@BenjaminHae thanks for catching that. I forgot about compiling 😬

PR #213, opened 2021-05-13
add option for hiding empty dashboard cards (3)

BenjaminHae:
The functionality discussed in #137 and #189 is completely implemented.

I've added tests for the cards and tried to add a test for the setting.
However, I did not find out how to correctly test the change of the setting (I've included my latest try in this PR).
@cdubz perhaps if you have time you can point me in the right direction?


babybuddy/tests/tests_forms.py
```diff
@@ -30,6 +30,7 @@ def setUpClass(cls):
             'last_name': 'Name',
             'email': 'user@user.user',
             'dashboard_refresh_rate': '',
+            'dashboard_hide_empty': 'off',
```
cdubz, line 33: FYI there have been some changes to the template settings here in recent commits. You may want to rebase once more just to make sure there are no conflicts. I don't think there will be any,

dashboard/templatetags/cards.py
```diff
@@ -9,24 +9,34 @@
 
 from core import models
 
-
 register = template.Library()
 
 
-@register.inclusion_tag('cards/diaperchange_last.html')
-def card_diaperchange_last(child):
+def _user_wants_hide(context):
```
cdubz, line 15: Kind of nitpicky but I don't like the `user_wants_hide` naming here. Can you refactor this and the related variables as just `hide_empty`?

dashboard/templatetags/cards.py
```diff
@@ -198,45 +256,56 @@ def card_statistics(child):
...
-        'type': 'float',
-        'stat': weight['change_weekly'],
-        'title': _('Weight change per week')})
+    if weight:
+        stats.append({
+            'type': 'float',
+            'stat': weight['change_weekly'],
+            'title': _('Weight change per week')})
+
+    empty = len(stats) == 0
```
cdubz, line 302: This _looks_ fine to me, but in my quick testing on your branch I never saw the statistics card -- it was always hidden even though there should have been statistics. Could you double check this functionality?

cdubz (changes requested):
A few change requests here but overall this is looking good! I wasn't able to get to your tests question (babies waking up now) so I'll try to take a look at that another time. Thanks for sticking with this!

babybuddy/tests/tests_forms.py
```diff
@@ -132,3 +133,13 @@ def test_user_settings_timezone(self):
...
+
+    def test_user_settings_dashboard_hide_empty_on(self):
+        self.c.login(**self.credentials)
+
+        params = self.settings_template.copy()
+        params['dashboard_hide_empty'] = "on"
+
+        page = self.c.post('/user/settings/', data=params, follow=True)
+        self.assertEqual(page.status_code, 200)
+        self.assertTrue(self.user.settings.dashboard_hide_empty)
```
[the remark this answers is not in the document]
BenjaminHae: This did not work, I've now also added a similar test for `dashboard_refresh_rate` which also does not work.
I guess this is associated with the test `test_user_settings_language` failing.

dashboard/templatetags/cards.py
BenjaminHae: I've renamed it.

dashboard/templatetags/cards.py
BenjaminHae: You are right, there was an error in the template.

babybuddy/tests/tests_forms.py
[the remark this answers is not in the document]
cdubz: It looks like this worked out fine. I just pushed a small change to fix `test_user_settings_dashboard_refresh_rate` (f714a9ab20c17f0365876ba37ec885e547398623). Both tests pass for me!

cdubz (approved):

cdubz:
Thanks @BenjaminHae and @jcgoette! This is a nice new feature. Will roll it in to a release soon.

PR #220, opened 2021-05-31
change ADD to COPY

jcgoette:
Changing ADD to COPY per [Dockerfile best practices.](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy)


cdubz:
Thanks, @jcgoette!

PR #221, opened 2021-05-31
update post: sections

jcgoette:
This should bring all API endpoints' post sections up to date with expected and actual API behavior.


cdubz:
Thanks, again!

PR #223, opened 2021-06-01
Feedings

### LINKED ISSUE #192: Feature/UI Request: Only require a single time for feedings
It would be nice if the UI would only require a single time for feedings, and if that time/date was auto-populated as current. I can see use cases where a duration of feeding would be useful, but for the most part, time and volume are all you really want to know. It's a small thing, but data entry at 2:15AM on a Wednesday is hard, so anything to simplify the process is nice. Similarly, feeding type (bottle, milk, ect) could be auto-populated from the previous entry, as this is usually consistent, and only a blurry-eyed volume entry was required.

Love the project. Thanks again.

jcgoette:
Closes #192.

Feel free to leave off  90e0921. I had just noticed, and agreed with, the second request:

>Similarly, feeding type (bottle, milk, ect) could be auto-populated from the previous entry, as this is usually consistent

Also, please double check for any potential issues with statistics in `dashboard/templatetags/cards.py` if `duration` == 0.

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -48,10 +48,10 @@ def set_initial_values(kwargs, form_type):
         last_feeding = models.Feeding.objects.filter(
             child=kwargs['initial']['child']).order_by('end').last()
         if last_feeding:
-            last_type = last_feeding.type
+            last_method = last_feeding.method
             last_feed_args = {'type': last_feeding.type}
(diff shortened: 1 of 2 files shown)


cdubz (approved):
Looks good. Thanks, @jcgoette!

PR #225, opened 2021-06-03
Update django.po

svenvdmeer:
I've read the guidelines and I was on able to figure out how to do the tests, I'm not experience enough with these things. I would however like to help with the dutch translation which has some serious problems. I have only changes the values of the translation so I hope this is still helpfull. If there is a better way to do this please let me know. I would also like to add some missing translations but I don't know how.

Dutch translation is inconsistent especially with the translation of the word diaper. Sometimes the word 'pamper' is used which is a brand. I've change the translation of 'diaper change' to 'luierverschoning' because the word 'verschoning' in itself is clearer then the word 'vervanging' if 'aanpassing' which is now used in the menu and at first I didn't understand what that was.

### DIFF
--- a/locale/nl/LC_MESSAGES/django.po
+++ b/locale/nl/LC_MESSAGES/django.po
@@ -125,7 +125,7 @@ msgstr "<strong>Fout:</strong> Enkele velden zijn foutief. Zie hieronder voor de
 #: babybuddy/templates/babybuddy/nav-dropdown.html:27 core/models.py:163
 #: core/models.py:167
 msgid "Diaper Change"
-msgstr "Luier vervangen"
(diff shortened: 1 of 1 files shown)


cdubz:
Hi @svenvdmeer -- thanks for contributing! There is [a note in the contributing documentation](https://github.com/babybuddy/babybuddy/blob/master/CONTRIBUTING.md#poeditor) about [POEditor](https://poeditor.com/projects/view?id=256149) which is a website with a much simpler interface for managing translations.

That said, your changes here look fine so I'll plan to build on this PR to get the improvements merged in. Thanks, again!

cdubz (approved):

Issue #227, opened 2021-06-25
`WORKER TIMEOUT` on Docker 20.x


lutzky:
1. Starting with the sample `docker-compose` files (either sqlite or postgres), set any `SECRET_KEY` and set `ALLOWED_HOSTS=*` (the latter is only required for gitpod, see below, but left here for consistency).
2. `docker-compose up`
3. Point the browser at the babybuddy instance, at the login page (or any other page) hit refresh 3-4 times. Most of the time it'll be sub-second latency, but occasionally it'll take about 30 seconds.
4. When this occurs, the logs will (after 30 seconds) say the following (with the numbers 13,14 incrementing each time)
    ```
    [1] [CRITICAL] WORKER TIMEOUT (pid:13)
    [13] [INFO] Worker exiting (pid: 13)
    [14] [INFO] Booting worker with pid: 14
    ```

Setting `DEBUG=True` and setting log-level to `debug` adds no information. This also occurs with https://github.com/linuxserver/docker-babybuddy.

Tested on (❌ means reproduces, ✔️  means doesn't-reproduce and works fine)

* ❌ Ubuntu 20.04.2 on WSL, Docker 20.10.6
* ❌ Ubuntu 20.04.2 on bare-metal, Docker 20.10.6
* ✔️ Ubuntu 20.04.2 on https://gitpod.io, Docker 19.03.15

This has been discussed in https://gitter.im/babybuddy/Lobby?at=60d4adb16c992105fd994cee, and I'm formatting it as an issue at the request of @cdubz.

I've performed some more investigation, I'll add in a comment below.

lutzky:
As to "why would the docker version be relevant":

* benoitc/gunicorn#1801 sounds similar and seems to indicate that this might be some kind of weird worker-tmp-dir handling
* Running `docker exec -it babybuddyapp ls -l /proc/${PID_OF_THE_WORKER}/fd` always shows a *deleted* `/tmp/wgunicorn-XXXXXXXX` file. That sounds like something that might confuse docker, or change behavior between docker versions.
  * This occurs on both docker-19 and docker-20, but it's conceivable that there's an unusual "read-after-delete" behavior that gunicorn relies on which would be docker-version-specific.
  * I've tried modifying `worker-tmp-dir` to a directory mounted on the host filesystem, but it doesn't seem to matter.

cdubz:
[...]

PR #228, opened 2021-06-25
Make docker-compose examples unicorn-appropriate

lutzky:
See https://pythonspeed.com/articles/gunicorn-in-docker/

Fixes https://github.com/babybuddy/babybuddy/issues/227


cdubz:
Thanks for your search and efforts on this, @lutzky!

PR #241, opened 2021-07-05
Reverted tempus dominus to the previous version to fix an issue with marking fields as read only

### LINKED ISSUE #224: Disable autocomplete for date and time fields
autocomplete
When I want to a date and time I get the tool that helps me select the date and time but on top of that chrome shows the autocomplete field which shows dates and times I've already added which is not useful at all.

This is easily fixed by adding autocomplete="off" to the form. Autocomplete could be useful for the notes section of the form but not really necessary.

ntrecina:
### Summary
It seems like a bug was introduced in version `5.19.0` of tempus dominus which is causing the readonly attribute to be stripped from the date/time input which causes the formatting to get messed up and you can interact with the input which causes issues on mobile devices. This reverts and locks it to the previous version. It looks like v5 of tempus dominus is no longer under development so maybe when v6 is released/stable we can see if the bug is fixed there.

This should fix #224.


cdubz:
Perfect -- thanks, @ntrecina!

Issue #229, opened 2021-06-26
Mobile UI Tweaks


lutzky:
Looking to use babybuddy instead of our current tracking app, I noticed a few things that I *think* are low-hanging fruit for UI cleanup, pertaining mostly to phone usability. (I'm making it a checklist here in hopes of it being a [newfangled task-list thing](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists#about-issue-task-lists); ideally the owner would turn their favorite suggestions into issues of their own)

- Improvements to timeline view
  - [x] #244
  - [x] #270
  - [x] #271
  - [x] #243 
  - [x] #240
  - [x] #247
  - [x] #272
  - [x] #246 
- [x] #239 
- [x] #251 
- Less vertical scrolling please
  - [x] #213
  - [x] #230
  - [x] #231
- No horizontal scrolling please
  - These are things that can be done to reduce horizontal scrolling in activity log pages.
  - [x] #235
  - [x] #236
  - [x] #232
  - [x] #237
  - [x] #233
- Less clicking please
  - [x] ~~Unified event log~~ this is the timeline view
  - [x] #234

BenjaminHae:
>   * [ ]  Dashboard: Hide any boxes that just say "None"
>     
>     * Or at least move them to the bottom
This was already implemented as part of #213, #215 additionally adds an option to hide old data.

cdubz:
👋 @lutzky. Thanks for these recommendation. UI/UX has never been my thing so I'm always happy to get ideas like this. I have converted a few that are clear enough to issues and have some questions about others:

> Dashboard: Change "Last thing" timeformat from "duration ago" to "HH:MM (duration ago)" 

I'm not totally sold on this. I've always liked the current time format as being very clear and easy to parse. I.e. there is no need to understand or learn that "02:45" translates to "2 hours, 45 minutes" because it is stated literally. I also wonder about the L10n implications of this change...

> Move Color and Amount columns to far right

Could you mock this up one up? I'm unsure what you're thinking here.

> Allow shorter timestamp formats

Where do you see this as a problem? I think the default used in e.g. lists is sufficiently short but maybe that's only because the current month names I'm looking at as example are short (July and June) (:

> Unified event log 

Not sure I follow this one. The child page itself has a "timeline" of sorts but it's probably not the best format. Is that the kind of thing you are thinking? It's also not really easily discovered. The dashboard is king.

lutzky:
[...]

PR #243, opened 2021-07-06
Timeline: Show diaper change details

lutzky:
Suggested in #229

### DIFF
--- a/core/timeline.py
+++ b/core/timeline.py
@@ -19,10 +19,18 @@ def get_objects(child, date):
     instances = DiaperChange.objects.filter(child=child).filter(
         time__range=(min_date, max_date)).order_by('-time')
     for instance in instances:
+        contents = []
+        if instance.wet:
+            contents.append('💧wet')
+        if instance.solid:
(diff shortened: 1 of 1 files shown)


core/timeline.py
```diff
@@ -19,11 +19,17 @@ def get_objects(child, date):
     instances = DiaperChange.objects.filter(child=child).filter(
         time__range=(min_date, max_date)).order_by('-time')
     for instance in instances:
+        details = ''
+        if instance.wet:
+            details += '💧'
+        if instance.solid:
+            details += '💩'
```
cdubz, line 26: Let's stick to words for now. I'm not against using emoji generally to convey information but I'm worried about the usability of this with teeny emoji and no other hinting. E.g. --

image

cdubz (changes requested):

core/timeline.py
lutzky: I find the emoji easy to help scan, so how about both?

cdubz (approved):

PR #247, opened 2021-07-08
Show duration in timeline

lutzky:
Suggested in #229


cben:
Screenshot_20210709-171622

This shows it both on start and end events.  Perhaps would make more sense to only show duration on the end event?

lutzky:
As we say at work, your slight preference is my demand 😁

PR #248, opened 2021-07-08
Dashboard: Move all "last" cards to top

### LINKED ISSUE #230: Dashboard: Move "Last Diaper Change" to near-top

lutzky:
Fixes #230

### DIFF
--- a/dashboard/templates/dashboard/child.html
+++ b/dashboard/templates/dashboard/child.html
@@ -12,15 +12,15 @@
 {% block content %}
     <div id="dashboard-child" class="card-columns">
         {% card_feeding_last object %}
+        {% card_diaperchange_last object %}
+        {% card_sleep_last object %}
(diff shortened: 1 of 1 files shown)


BenjaminHae:
This might be cool on mobile but on a desktop the "latest" cards are on the left. I think this is a bit confusing.

lutzky:
The horizontal positioning on desktop is arbitrary anyway due to "Hide Empty Dashboard Cards".

BenjaminHae:
Only when the cards are empty.

cdubz:
@BenjaminHae yeah considered that but I think it is OK. It would be nice to have them across the top, I think. But having them in a row along the left side also make sense to me as a “most recent information is here” setup. I’ve had a looooooong term plan to make their position fully configurable with a drag-n-drop library (there is an issue somewhere) but have never got around it. Maybe it’s time to revive that idea…

PR #249, opened 2021-07-08
Add a Submit button at the form top

### LINKED ISSUE #231: Don't hide Submit button below optional fields

lutzky:
Fixes #231

### DIFF
--- a/babybuddy/templates/babybuddy/form.html
+++ b/babybuddy/templates/babybuddy/form.html
@@ -3,6 +3,7 @@
 <div class="container-fluid">
     <form role="form" method="post" enctype="multipart/form-data">
         {% csrf_token %}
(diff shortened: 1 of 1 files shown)


cdubz:
@lutzky I don’t love the look of this. To me it seems to make the forms look confusing or maybe broken.

I’m thinking a better approach would be float the bottom submit button such that it sticks at the bottom during scroll. That or maybe have the form page header float with the menu bar and put the button in there.

Both of those options are a bit more complicated. I’ll try to find some time to give it a try if you don’t get around it.

lutzky:
Understood. This requires stronger CSS-foo than I can quickly whip up though, so I guess this can be on the back-burner.

cdubz:
Closing in favor of #254 (with some incoming changes).

PR #251, opened 2021-07-08
Brighten blue and cyan colors

lutzky:
For slightly easier viewing on mobile phones in daylight

Not mentioned in #229 but closely related.


cben:
Could you add a screenshot?
@cdubz your demo app runs on heroku, right? Consider enable Review Apps option in heroku, you should get live samples to play with for all PRs ✨, it worked on my fork.

Looking at dashboard, but without testing:
Cyan on black is already pretty good but will benefit 👍 
Blue on grey is not so readable now, this should help a lot 👍 👍 
But there are also white on blue headings, whose contrast will become slightly worse?  Probably still OK.

lutzky:
Before:

image

After:

image

Agreed that white-on-blue is slightly worse, but I think it's more important for the content (which is smaller to boot!) to have good contrast.

+1 to Review Apps, sounds useful!

cdubz:
@cben @lutzky I do have review apps set up but I don’t autobuild them because Heroku doesn’t have any cost controls and it’s my personal credit card tied to the account. I generally only build them when I am doing my own review.

Anyway I’ll build them for the current group of PRs. I may be slow to review myself because of travel this weekend.

cdubz:
@lutzky you need to rebuild and commit the static files as well for this change to take. I believe it’s just `gulp updatestatic` that is needed.

cben:
I just played with this vs regular demo on my phone outside, agree it's a definite improvement!

Some barely readable light-grey-on-gray raint but not a blocker to merging this. 🚀 

Oh, i see, i thought they can go on free plan.  Added a small sponsorship fwiw.

cdubz:
Thanks @cben! I went ahead and enabled automated builds. Should be fine as long as no spams the repo with lots of PRs and keeps them alive.

cdubz:
Anyway I’m on board for this change. The contrast is way better. Thanks @lutzky!

PR #254, opened 2021-07-14
Make form submit full width fixed

### LINKED ISSUE #231: Don't hide Submit button below optional fields

cdubz:
Closes #231. Alternative to #249.

@lutzky @BenjaminHae @ntrecina interested in your thoughts on this approach (since y'all chimed in on #249 to some extent). Mobile experience seems solid but I'm not sure how I feel about desktop -- the ux is not better there but I also don't think it is _worse_.


lutzky:
I think it's slightly worse on desktop; ideally this would only occur on mobile, but I think it's more complicated to get that to work within bootstrap, and not necessarily worth it, so I'm in favor of this change.

cdubz:
I don’t think it would be much effort to make this more responsive. This was just the lightest touch approach. I’ll sit on it for a moment.

ntrecina:
@cdubz I like the ux on mobile but I'm also on the fence about desktop. I don't think it's terrible but it does feel a bit off. I think if you could split the experience between mobile and desktop that would be perfect.

Issue #217, opened 2021-05-26
Proposal: add generic custom events (that users can define and track)


philip-n:
Browsing through the issues, there are multiple requests for specific additional tasks/things to keep track of (medication, baby height, ...).

Most of them have almost the same requirements:
* a name of the task
* some default fields (pretty much always identical)
  * start date (or simply: date)
  * optional end date
  * note
* some more custom-y fields that can be relevant
  * a quantity
  * potentially additional convenience features:
    * multiple choice / dropdown style selection between predefined values
    * named boolean values that are displayed as toggle-buttons (to allow quick input like `fluid` vs. `hard` with the diaper change)

Rather than implementing explicit models for all of the requested features, I'd propose adding a generic `CustomEvent` model that offers the generic fields listed above. The user could then first add the tasks they want to track in an admin menu and potentially define which of the available fields are relevant. These tasks would then be available in the frontend as usual, maybe under a new "custom tasks" item.

In my opinion, this would significantly extend the functionality (what you can jot down on a piece of paper, you could then also track in Baby Buddy) without bloating the interface with various things that some users might never care about (as it only offers what you configured).

Also, having everything in the same model would make it easy(-ish) for example to extend the sleep pattern style view to display additional tracked things.

Thoughts? Opinions?

philip-n:
To add some links to relevant issues:
* Track other baby care tasks (bathing, nail trimming...) https://github.com/babybuddy/babybuddy/issues/176
* Ability to track Medicine https://github.com/babybuddy/babybuddy/issues/153
* Feature Request: track milk pumping https://github.com/babybuddy/babybuddy/issues/159
* Feature Request: Vitamin and Medicine Tracking https://github.com/babybuddy/babybuddy/issues/66

cdubz:
This is an interesting idea. I certainly would not be opposed to it. This is kind of how I envision the "Notes" model right now so maybe we could just expand on that...

If this generic model itself is configurable (as you outline in the "some more custom-y fields that can be relevant" list) that will definitely ramp up the complexity both from a maintenance and UX perspective. It would be nice if we could, as you say, identify some of these common factors and then maybe add just _one_ system-wide configurable field "category" (or something) that would be a select list on the generic.

For example a user could configure categories of "Bathing", "Nail trimming", "Pumping", etc. and would select one of those options for the entry. That would probably make it easier to pivot data for things like display, data viz, and API retrieval.

So if we wanted to just repurpose notes for this we could:

- Add a title field
[...]

PR #257, opened 2021-07-26
I wanted to keep track of what songs my daughter likes/dislikes 

mknod:
I wanted to keep track of artists/songs i used to help soothe my daughter. I mostly used the already provided templates and models here and tried to stay consistent. Let me know if there are improvements that can be made. Thank you!


cdubz:
@mknod thanks for submitting such a thorough PR. While this work looks good, I have been avoiding merging new core models like this for a while now because the list of possible ones is seemingly endless. In #217 we are discussing how we might go about adding a generic, customizable model for folks to use. Any collaboration on that front would be greatly appreciated.

PR #260, opened 2021-07-31
Use short datetime string for lists

### LINKED ISSUE #237: Use shorter timestamp formats in list view

cdubz:
Closes #237


cdubz:
@lutzky curious to get your review on this if you have a moment. The date still gets broken up in to multiple lines on small screens but the other recommendations in #229 should help more as they land.

lutzky:
Well, this seems strictly better. Dropping the year makes the biggest difference, and then a three-letter month name makes more sense than numeric. "Today" is a nice touch.

PR #262, opened 2021-08-02
Hide child column if only one child is shown

lutzky:
Show child name in heading instead

WIP for #233 for review; currently only diaper changes implemented


cdubz:
@lutzky this looks good to me. I’d say we don’t even need to add the child name to the header. If there is only child in the system that context seems unnecessary.

PR #267, opened 2021-08-04
Move Actions column to the far left

### LINKED ISSUE #236: Move Color and Amount columns to far right in list view

lutzky:
Fixes #236


cdubz:
@lutzky seeing this now I'm actually wondering if it would be better to just have the "Actions" row _first_ in all list views... I feel like going to list views to delete or edit entries is probably the most common use case (since dashboard delivers all the other valuable information). Thoughts?

lutzky:
I like it. Shall I do that instead of this?

cdubz:
Yeah -- lets do that instead.

PR #268, opened 2021-08-04
Add links from dashboard

### LINKED ISSUE #234: Links from dashboard

lutzky:
Fixes #234


cdubz (changes requested):
@lutzky insce you made changes to a sass file you'll need to also run `gulp updatestatic` and commit the changes.

lutzky:
*sigh* right; would be nice for the static files not to be in the repo in the first place.

cdubz:
Yeah I know. I do that so that all the node dependencies don't have to be part of the deploy chain. For e.g. manual deploys in low memory situations it's a pain in the ass but maybe that's not really relevant this days...

I'd also love to have GitHub stop showing the primary language for this repo as JavaScript hah.

lutzky:
I mean for *docker* that shouldn't be a problem, using [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) (so the dev dependencies are only installed in the build stage, but the final image doesn't have them), but I don't know about the other deployment styles. :shrug:

lutzky:
#269 should help with the repo language

cdubz (approved):

PR #285, opened 2021-08-21
add SECURE_PROXY_SSL_HEADER env arg option

jcgoette:
Adds option to use Django's SECURE_PROXY_SSL_HEADER option. 

Default to null. Readme bears same warning as Django's docs.

Documentation per https://docs.djangoproject.com/en/3.0/ref/settings/#secure-proxy-ssl-header.

I don't think any tests can be added?

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ take precedence over the contents of an `.env` file.**
 - [`DB_PORT`](#db_port)
 - [`DB_USER`](#db_user)
 - [`SECRET_KEY`](#secret_key)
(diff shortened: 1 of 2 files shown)


cdubz:
Thanks @jcgoette! Technically there are some other non-standard forms of this header that we _could_ support (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto#examples) but we can address that if it ever affects an actual user...

jcgoette:
Interesting, didn't know. We could pass the var value into the first piece of SECURE_PROXY_SSL_HEADER tuple to account for non-standard headers. But I'm ok waiting for an actual user also. 😊

Issue #289, opened 2021-08-21
Sleep graph y-axis should go "down" from midnight at the top.


cdubz:

cdubz:
Closed by #295.

PR #295, opened 2021-08-28
reversing y-axis of sleep plot (top to down)

fodi666:
should be enough to solve #289


cdubz:
Taking a quick look on mobile — I think some additional adjustments (to either the axes ranges or calculations) are still needed. See e.g. —

FE92670B-B401-4E46-94AC-A137D9427A75

fodi666:
I checked on my mobile and it was not that bad for me. However, I realised that I don't need the "autorange" option, instead it is enough to define the range in reverse order. (note: the plot does not display the last label even with the default settings)
image
image

cdubz:
Nice. That looks like it addresses it!

PR #303, opened 2021-09-03
Gitpod config

lutzky:
This lets contributors easily start a one-click dev environment for their patches, including all necessary software. Not that I'm being lazy, but I'm being lazy.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,13 @@
+tasks:
+  - init: >
+      pipenv install --three &&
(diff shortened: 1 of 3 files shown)


cdubz:
Interesting -- I had no idea this was a thing! The badge should go in `CONTRIBUTING.md` _or_ we should move those (CI and coverage) badges back to `README.md`. I think I lean towards moving them back to `README.md`. @lutzky could you add that change to your branch?

Issue #284, opened 2021-08-20
Allow manual editting of datetime values


Expro:
Please restore ability to correct time entry from keyboard by deleting and writing down numbers, instead of several clicks through GUI. Baby can be demanding and requiring attention, so slowing down ability to add correct entry is unwelcome.

cdubz:
@Expro this is actually intentional and the previous behavior was treated as bug (see #224 and comments). Are you saying to prefer to be able to click in to the date/time field and manually change the text after making a selection with the date picker? If so, why? Just wondering if there is also a datepicker UX issue to be considered here.

Expro:
> If so, why?

This are scenarios where it comes handy:

- App was inaccessible for some reason, and I have noted time as text somewhere else. With date and time field as text field I can just dismiss datepicker and copy-paste in time. In newer version, I have to go through GUI and recreate time value, which is slower and less comfortable.
- I have started timer with some delay (usually with 2 / 3 minutes late, cause baby comes first), so I just want to correct time value. With time being text, I can just delete 2 last numbers. With GUI only, I have to switch from date to time and then do several clicks to fix time value.
- Text field allows me to use speech to text to dictate time. 

> Just wondering if there is also a datepicker UX issue to be considered here.

There is definitely one thing that could be improved - datepicker opens with view focused on date, rather than time, and it should be reversed. I rarely have to provide time different than "today" (either when putting in old notes, or on action that started yesterday, but ended today) but I provide time many times per day. Please optimize default view for more common action.

Other than that, date / time picker is very well done.

cdubz:
Ok. That all makes sense. Interesting use case with speech text!

Expro:
I think there is enough space even on mobile screens to layout with button for date picker, button for time picker and textbox with resulting date + time that is editable (without triggering pickers when focused). That would allow user to select if he wants to pick date, time, or edit it by hand.

cdubz:
Interesting idea. Not sure if the existing picker library BB uses supports that but I have also been looking for excuses to move away from it because it’s had lots of weird refactoring over the years. Though every time I have looked I haven’t really been able to find a comparatively good picker.

cdubz:
@Expro I’d be interested in your thought on #304 related to this.

PR #304, opened 2021-09-06
Date picker changes

ntrecina:
### Summary
I also noticed a regression from 0a74061a7aedc14734735759cce46d6d5aafc392 where the readonly attribute was removed so mobile browsers were once again displaying the keyboard when tapping on the date picker field. I added the `inputmode=none` attribute which seems like it might be a better solution while still allowing the field to be editable via a keyboard if desired.


cdubz:
@ntrecina what do you think of the discussion in #284? I think it’s reasonable to allow direct modification of the datetime fields for certain use cases and the UX is not necessarily bad with the keyboard appearing on mobile.

cdubz:
Re: the static files — could you see if you get the same result from `gulp updatestatic`? I don’t _think_ we should be generating FA fonts anymore. Unless they are used in Django admin?

ntrecina:
@cdubz I think it makes sense to have direct modification of date/time fields as mentioned in #284 as I have run into some of the scenarios mentioned. However, I may be wrong but I think the 99% use case on mobile will involve using the gui picker. The keyboard popping up every time you adjust a date/time is a very frustrating user experience. If we were able to come up with a compromise where by default on mobile it doesn't show the keyboard unless you manually chose to I think that would be ideal but as you mentioned I'm not sure if that's possible with the current library.

I'll give `gulp updatestatic` a try, thanks!

cdubz:
@ntrecina could you open a new pr with just the breast milk autoselect change? That seems fine to me but I'd like to invite more discussion about the `inputmode` change.

ntrecina:
@cdubz Sure thing.

ntrecina:
@Expro Do you have any thoughts on this change? From reading your comments on #284 it sounds like you would like to modify date/times via the keyboard on desktop. Would you also like that ability on mobile or do you think gui only control on mobile would be ok?

Expro:
I'm using BabyBuddy almost exclusively on mobile devices, so all my comments are valid for mobile. I still think separate button for GUI picker invocation that is placed adjacent to textboxt containing date and time is superior solution.

cdubz:
@Expro I see -- you are advocating for not showing the _picker_ when clicking in the _field_ but only when clicking the little calendar icon next to the field, right? I'm not sure to what extent that behavior is configurable with the library but that approach does seem like a good middle ground if we can sort it out.

@ntrecina what do you think?

Expro:
That right - I think editable field AND ability to launch picker by button / calendar icon covers both use cases intuitively and would be great if possible (as You have mentioned).

ntrecina:
That seems like a good compromise to satisfy both use cases. That being said I would still prefer an option (user preference maybe) to block keyboard input on mobile since my wife and I don't use it and it's nice having a large target to tap on rather than needing to tap on the calendar to avoid the keyboard opening.

cdubz:
Fair enough, thanks for contributing to the discussion @ntrecina and @Expro. I'm going to merge this change and we'll see how it goes. I think I need to put some more serious thought in to settings organization and options so that I am less overall adverse to adding toggles like this.

PR #307, opened 2021-09-09
Feeding form type breast autoselect

ntrecina:
### Summary
Per a request from my wife I've added a small feature to the feeding form that will auto select breast milk as the type when choosing a breast option as the feeding method.


cdubz:
Thanks!

Issue #283, opened 2021-08-19
Sleep graph issues


fodi666:
- [x] #286
- [x] #287
- [x] #288
- [x] #289
- [x] #290

--

Hi,
I have found that the sleep pattern plot is misaligned. I attach an image with the problem.
![newplot](https://user-images.githubusercontent.com/1128180/130141653-02077737-1ea5-439f-8ed0-bcba164a3ff1.png)

The first column is supposed to be 17 August while the second column of sleep entry is today (19 August), with no entries from 18 August. Yet, the two days are connected, so it seems that the patterns are valid for two days, but most importantly, the times don't match between the two columns. The first column seems to be at the good position, while today's plot is at weird time slots.
Actually the first problem (weird x-axis representation) is also present on other graphs, for example the feeding graph:
![feeding](https://user-images.githubusercontent.com/1128180/130141996-9344cf5b-6afd-4bf6-9448-94fbed3fab69.png)

I think that the columns should range from 0:00 to 23:59 each day.

Also, I'd have a suggestion with this: could you invert the y-axis, so the day goes from top to bottom instead of bottom to top? It would make it easier to follow.
Thanks

cdubz:
Hm yeah... I'm guessing that somewhere along the line we assume that entries for the sleep pattern graph do not have any gaps. That's definitely a bug of some sort.

@fodi666 re: the other parts of your comments on weird x-axis representation and inverting to y-axis -- I'm not totally clear what the ask/desired functionality is. Could you clarify? Or maybe rough up an example graph of what you are thinking?

fodi666:
OK, so it seems that the missing day gets compensated once I had some more days added and now it honors the missing day correctly.
However, Aug 19 is still shifted, if you check the screenshot, the bottom sleep for Aug 19 is 1258 AM to 0350 AM, yet it is lined with a sleep from the other day at 1053 AM to 1230 PM. This for some reason stayed. I actually have a 24h setting with environmental variables set and the UI uses 24h, however this graph for some reason uses 12h format, which is also strange.
The weird x-axis is the fact that it is shifted by 12 hours if you check the columns and the x axis labels. This is also visible on the other plots (the start of the day on the axis is actually in the middle of the columns).

![Inkedsleep pattern bug](https://user-images.githubusercontent.com/1128180/130329330-6163ea5c-78fc-455f-b843-994194893aa6.jpg)

[...]

PR #326, opened 2021-10-30
refacto(sleep-reports): #283 Rename some variables and add constants for magic strings

codisart:
Rename instances/instance to sleeps/sleep to facilitate identification of concepts
Add constants to name some magic strings

### DIFF
--- a/reports/graphs/sleep_pattern.py
+++ b/reports/graphs/sleep_pattern.py
@@ -11,23 +11,26 @@ from core.utils import duration_string
 
 from reports import utils
 
+ASLEEP_COLOR = 'rgb(35, 110, 150)'
+AWAKE_COLOR = 'rgba(255, 255, 255, 0)'
 
-def sleep_pattern(instances):
+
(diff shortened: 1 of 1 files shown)


reports/graphs/sleep_pattern.py
```diff
@@ -50,7 +52,7 @@ def sleep_pattern(instances):
         if end_time.date() != start_time.date():
             adj_start_time = end_time.replace(hour=0, minute=0, second=0)
             adjustment = {
-                'column': end_time.date().isoformat(),
+                'column': end_date,
```
cdubz, line 55: @codisart Are you sure this change (removing the `date().isoformat()` has no downstream effects? Otherwise this PR all looks fine.

codisart: I am not sure but end_date is init with ` end_time.date().isoformat()` at line l.29.

cdubz:
Looks good :+1:

PR #327, opened 2021-10-30
refacto(sleep-reports): #283 Replace times and labels lists by a single days list

codisart:
I was thinking to try to put together the times and the labels in a dictionary.
I think it clarifies a little taht they go together. Maybe the colors could go in the dict.
It could allow to add napping behavior later.

### DIFF
--- a/reports/graphs/sleep_pattern.py
+++ b/reports/graphs/sleep_pattern.py
@@ -112,8 +112,8 @@ def sleep_pattern(sleeps):
         text = {}
         for date in days.keys():
             try:
-                y[date] = days[date][i].time
-                text[date] = days[date][i].label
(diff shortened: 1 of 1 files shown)


cdubz:
@codisart I think this looks good except for the minor linting issues. Will take a look once those are resolved :+1:

codisart:
So I am not sure what the reason for the failed test. I would investigate

cdubz:
@codisart accessing collection members uses brackets in Python (as opposed to dots for accessing attributes and methods). So the test was failing because the sleep patter graph page was not working (:

I fixed that on your branch. Tests looking good now and again this refactor is nice help on readability -- thank you!!

PR #332, opened 2021-11-11
Update format customizations handling

### LINKED ISSUE #324: Custom `en` date formats interfere with `en-GB` language
@cdubz Thanks for trying to get #321 sorted.

I've had a quick test of the latest commit (fae764a) and it wasn't fully working for me. (This should probably be moved to an  issue)

Data input is working correctly and the input controls use the correct format, but the display list pages are not.
Not sure if this is expected or not (from a tech perspective), but the default time format when en-GB is selected remains 12h not 24h.

It seems that `Timeline` and `Children` are displaying correctly, but the following are not:

`Notes`
`Temperature`
`Weight`
All of the `Activity` list views.
`Timer` list view.

The odd thing was `Timer` (detailed view) was showing the date correctly (albeit with the default 12h time format), but when switching using `USE_24_HOUR_TIME_FORMAT=True` the time format was correct but not the date.

I was using python 3.8.12 if that makes any difference.

_Originally posted by @Amith211 in https://github.com/babybuddy/babybuddy/issues/321#issuecomment-952748371_

cdubz:
@Amith211 :wave: -- I'd appreciate your feedback on this branch. This moves the format handling to middleware because there seems to be no agreeable way to make this work with format modules.

Re: the 12h vs. 24h display -- Django's [default formats for `en-GB`](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/conf/locale/en_GB/formats.py) actually use 12h format. The `USE_24_HOUR_TIME_FORMAT` setting needs to be implemented per-locale when overriding the defaults. Do you think that should be necessary here? I'm not sure what the standard is in the UK but I'd be surprised if Django's defaults are not the norm.

Fixes #324.


cdubz:
I'm pretty confident in this approach so I'll move it forward (want to move a 1.9.1 release but not with the en-GB unhappiness). Will address any issues in further tickets.

Amith211:
@cdubz Apologies, I didn't get a chance to have a look the other day.  

I'm not familiar enough with Python/Django to make any strong comments regarding implementation; I mainly using c#. For my understanding, would the overrides follow the same partern here, requiring a `update_{locale}_date_formats()` function to implement `USE_24_HOUR_TIME_FORMAT` for that lang?

I'll open a new issue to address the 24h format for `en_GB`. The short version is that for `en_GB` `USE_24_HOUR_TIME_FORMAT` needs to be implemented as this would be the expected format when shown on a digital display for most, if not all, Brits.

cdubz:
No worries! I'm just antsy for a release (:

So you for customizing a particular locale you could pretty much follow exactly what was done for "`en_US`" in this PR. I figured 24H time made more sense for y'all and I find it very odd that Django doesn't use that as its own format default for `en-GB`.

Amith211:
Yeah, odd they don't seem to follow the 'modern' standard, 12h is more common in spoken communication, but not for written communication, especially when formal or digital.  I'll see you over at #333 so I don't get confused :)

PR #335, opened 2021-11-13
add en-gb datetime overrides

Amith211:
WIP, I've rushed this so I may have overlooked something, I've done a quick tests with both 12 and 24h and it seems to be working.


cdubz:
@Amith211 don't worry about that CI fail on the "deploy" action. Thought I had that property configured to _not_ run on PRs but apparently not... (:

Amith211:
@coveralls  Please could you take a look to make sure I'm on the right track :)

It seems to be working save for a couple of bugs I can't get my head around.

1. When `USE_24_HOUR_TIME_FORMAT=False` and `en_GB` are set the date time picker remains 24h
2. When switching between US and GB the format on `/changes` get stuck on the US format, until server restart

babybuddy/middleware.py
```diff
@@ -49,7 +78,8 @@ def __call__(self, request):
         user = request.user
         if hasattr(user, 'settings') and user.settings.language == 'en-US':
             update_en_us_date_formats()
-
+        elif hasattr(user, 'settings') and user.settings.language == 'en-GB':
```
cdubz, line 81: @Amith211 instead of doing the same check twice here, move both of the `user.settings.language` checks under a single conditional for `hasattr(user, 'settings')`.

babybuddy/middleware.py
```diff
@@ -38,6 +39,34 @@ def update_en_us_date_formats():
...
+        # These formats are added to support the locale style of Baby Buddy's
+        # frontend library, which uses momentjs.
+        custom_input_formats = [
+            '%d/%m/%Y %I:%M:%S %p',  # '25/10/2006 2:30:59 PM'
+            '%d/%m/%Y %I:%M %p',  # '25/10/2006 2:30 PM'
+        ]
+
+    # Not setting this makes pages display using the the 25/10/2006 as desired
+    # Add custom "short" version of `MONTH_DAY_FORMAT`.
+    # formats_en_gb.SHORT_MONTH_DAY_FORMAT = 'j M'
```
cdubz, line 63: @Amith211 this is optional. I do it specifically got en(-US) because the normal `MONTH_DAY` format uses full month names. If en-GB doesn't need the same treatment you can go ahead and remove this (and the relevant test).

babybuddy/tests/formats/tests_en_gb.py
```diff
@@ -0,0 +1,74 @@
...
+            try:
+                result = field.to_python(example)
+                self.assertIsInstance(result, datetime.datetime)
+            except ValidationError:
+                self.fail('Format of "{}" not recognized!'.format(example))
+
+        with self.assertRaises(ValidationError):
+            field.to_python('invalid date string!')
+
+    # @tag('isolate')
```
cdubz, line 35: @Amith211 I think we don't need to isolate these tests anymore. So you should be able to safely remove this tag and import.

cdubz (commented):
Code is all looking good -- you're definitely on the right path here.

cdubz:
@Amith211 and following up on these items --

>     1. When `USE_24_HOUR_TIME_FORMAT=False` and `en_GB` are set the date time picker remains 24h

This is because Moment (unlike Django) uses 24H as the default format for en-GB. The primary reason I implemented `USE_24_HOUR_TIME_FORMAT` was to _override_ 12H format locales. In that way the env name is kind of a misnomer. It should be, like `USE_24_HOUR_TIME_FORMAT_FOR_12_HOUR_TIME_FORMAT_LOCALES` 😄

I'm not 100% sure how I want to resolve this... we could change the moment formatting handling to do the _inverse_ but that would be weird/unexpected behavior for locales that are correctly configured for 24H by default. Maybe we should just use this as an excuse to go ahead with the deprecation of that env var and adding 12/24H as a user level setting.

Although maybe a less abrasive approach would be to handle that value as an inverse for en-GB only. To do that we'd have to get the request and user context in to [`core/templatetags/datetime.py`](https://github.com/babybuddy/babybuddy/blob/master/core/templatetags/datetime.py).

>     2. When switching between US and GB the format on `/changes` get stuck on the US format, until server restart

I wasn't able to reproduce this. Can you get it to happen consistently?

cdubz:
FYI @Amith211 I made some changes to the language handling generally that may improve things a bit here. I also added special handling for en-GB + 12H to enforce AM/PM for momentjs doesn't get confused.

I don't love this solution but I'm not sure there is another alternative outside full deprecation of the server-based 24H setting. I _want_ to do that but it seems a bit daunting...

Amith211:
@cdubz  Sorry I haven't had that much fee time lately to fully test this.  From a brief test the other day it seems to be OK. I'll have a another check this is evening. 

I was still having the weird changing issue above, but I can't fully narrow it down to either requiring clearing cached cookies or a server restart; it seems to be both an arbitrary cause and fix. 

Do we have a dockerfile for testing local changes? I had a quick try with copying local files to the LEO dockerfile but the server wasn't displaying for some reason. I want to try to eliminate any weird environment behaviour. I have been running the server in WSL when testing.

cdubz:
Oh and no worries @Amith211 on time to test. I'm going to ahead and merge this and we can create a separate issue for the changes problem if it persists. I think this plus one additional change will get https://github.com/babybuddy/babybuddy/issues/337 resolved for folks...

Amith211:
@cdubz Makes a lot of sense to get this and the other change in now as people are actually suffering with #337

I'll do some more testing of master and I'll also have another attempt at a dockerfile for local testing, if only a rudimentary version.

PR #338, opened 2021-11-16
Complete Spanish translations plus some fixes

Alberdi:
I've used the POEditor project to update them, this commit imports the changes directly from there.
The changelog looks dramatic but I swear I've only changed a few strings here and there.


cdubz:
Looks great. Thanks for contributing!

PR #341, opened 2021-11-17
fix(sleep-reports): #286 Init all days in the period to remove gaps

### LINKED ISSUE #286: Sleep graph has incorrect positioning when there is a gap in days.

codisart:
Fixes #286


cdubz:
@codisart check out the [linting](https://github.com/babybuddy/babybuddy/runs/4243422994?check_suite_focus=true) -- there appears to be issue with the `_new_day` method you added.

PR #344, opened 2021-11-20
Add notes to child timeline

### LINKED ISSUE #343: Notes on child timeline
The child timeline, as it appears in `/children/child-name-slug`, doesn't currently display the child's notes. I think adding them would allow to have a full summary of all the day events.

Alberdi:
I went with a simple approach that doesn't require any new translations and I think that is clear enough, but feel free to propose improvements.

Closes #343.


cdubz:
Looks perfect. Thanks!

PR #356, opened 2021-12-15
Convert envir variables that are supposed to be boolean to boolean

### LINKED ISSUE #354: Boolean environmental variables
Hi,

I'm running babybuddy via the linuxserver docker container and failed to turn off the debug mode via environmental variables. Note that the linuxserver docker container sets DEBUG to False if not set, but this is interpreted as a string and thus treated as True.

Some investigation showed that the dotenv package does not support boolean variables, and thus whatever non-empty value the DEBUG env variable takes, it evaluates to True. See this issue: theskumar/python-dotenv#86 for reference
A potential fix could be converting the strings that are supposed to be boolean into a proper boolean, similar to [how it's done for the 24h format](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/settings/base.py#L198), i.e.
`DEBUG = bool(strtobool(os.getenv.get('DEBUG','False')))`

Let me know if you would be willing to accept a pull request on this issue.

MagiX13:
dotenv does not take allow for boolean environment variables and thus they have to be converted before
This should fix #354

### DIFF
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -22,7 +22,7 @@ load_dotenv(find_dotenv())
 
 ALLOWED_HOSTS = os.environ.get('ALLOWED_HOSTS', '*').split(',')
 SECRET_KEY = os.environ.get('SECRET_KEY') or None
-DEBUG = os.environ.get('DEBUG') or False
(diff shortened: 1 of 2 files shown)


cdubz:
Perfect. Thanks for contributing, @MagiX13!

Issue #186, opened 2021-01-17
Host in a subfolder / different root path


friesenkiwi:
Hi,
instead of hosting on the root of a domain (like `https://example.com/`), I wanted to host babybuddy in a different path like `https://example.com/examplepath`.

I was able to almost fully achieve it by setting

```
FORCE_SCRIPT_NAME = '/examplepath'
STATIC_URL = 'examplepath/static/'
WHITENOISE_STATIC_PREFIX = 'static/'
LOGIN_REDIRECT_URL = '/examplepath/'
LOGIN_URL = '/examplepath/login/'
LOGOUT_REDIRECT_URL = '/examplepath/login/'
```

at the end of `production.py`.

There are only two minor glitches: All functionality and links work nicely, with the exception of the textual "Home" link and the link on the Babybuddy logo, which leads to https://example.com/ instead of https://example.com/examplepath/

What would be the best way to add that last bit of functionality?
* Add something like `    path('', ?, name='home'),` at https://github.com/babybuddy/babybuddy/blob/master/core/urls.py#L9
* Change `href="/"` at https://github.com/babybuddy/babybuddy/blob/master/babybuddy/templates/babybuddy/nav-dropdown.html#L6 and https://github.com/babybuddy/babybuddy/blob/master/babybuddy/templates/babybuddy/base.html#L36
?

I'm unfortunately not very literate in Python, otherwise I would have tried myself. But if you can give me some hints in the right direction, I'd put up a PR for it.

cdubz:
Hmmmm. It definitely seems like this should be easier than all that. I will take a look at this soonish. In the mean time, is a subdomain not an option you could use?

friesenkiwi:
Well, if you want to e.g. host locally on a Raspberry for privacy reasons, you'd have http://raspberrypi.local and it's hard to setup a subdomain. But you might want to host multiple services, so a subfolder seems the natural choice (at least to me…)

cdubz:
True. But in that case you could also just assign any other local domain you want to the Pi (in additional to the default one) — e.g. babybuddy.local. Regardless, I’m not judging your choice here (:  Just not sure when I’ll get a chance to test this.

friesenkiwi:
Well, if you want to e.g. host locally on a Raspberry for privacy reasons, you'd have http://raspberrypi.local and it's hard to setup a subdomain. But you might want to host multiple services, so a subfolder seems the natural choice (at least to me…)

friesenkiwi:
[...]

PR #358, opened 2021-12-17
Minor changes to run into subdir

MagiX13:
These changes should be sufficient to run babybuddy in a subdir if the relevant reverse proxy passes a SCRIPT_NAME to django (not HTTP_SCRIPT_NAME). This might be a django bug that I need to investigate a bit more.
In addition, this also needs a static/babybuddy entry in the relevant reverse proxy for now


cdubz:
Regardless of #186, these changes are good. Thanks!

PR #361, opened 2021-12-29
Optional last name

Alberdi:
I find it a bit detached to have the children referenced by their full name, specially because we use two last names in my country and we just use Baby Buddy at home. In my opinion, allowing to name it with just one word makes it feel more personal and less like a teacher doing a roll call.


cdubz:
Makes sense. Thanks for the contribution!

PR #367, opened 2022-01-10
Add timer restart and stop triggers to REST-API

MrApplejuice:
# Problem

I am writing an Android app right now where I want to leverage starting and stopping of timers on baby buddy to allow parents (typically two) to conveniently and easily log sleep/feeding/belly times from their phone without navigating through the website.

However, quite late in the development I figured out that the "active" flag is readonly in the REST-API. Also, even if writable, the handy feature where the end-time is marked by the server would not have been available if "active" was writable. The restart/stop triggers used by the form of the website are not available in the REST-API.

# Solution

This PR adds the endpoints `api/timer/{id}/[stop|restart]/`-triggers that can be PATCHed through the REST-API to restart/stop timers.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -160,6 +160,7 @@ management commands.
 - [`gulp docs:watch`](#docswatch)
 - [`gulp extras`](#extras)
 - [`gulp fake`](#fake)
+- [`gulp generatescheme`](#generatescheme)
 - [`gulp lint`](#lint)
 - [`gulp makemessages`](#makemessages)
 - [`gulp makemigrations`](#makemigrations)
@@ -236,6 +237,11 @@ Adds some fake data to the database. By default, ``fake`` creates one child and
 31 days of random data. Use the  `--children` and `--days` flags to change the
 default values, e.g. `gulp fake --children 5 --days 7` to generate five fake
 children and seven days of data for each.
+
+#### `generateschema`
+
+Updates the [`openapi-schema.yml`](/openapi-schema.yml) file in the project root
+based on current API functionality.
 
 #### `lint`
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -305,9 +305,6 @@ gulp.task('makemigrations', function(cb) {
     _runInPipenv(['python', 'manage.py', 'makemigrations'], cb);
 });
 
-/**
- * Runs the custom "reset" command to start a fresh database with fake data.
- */
 gulp.task('reset', function(cb) {
     _runInPipenv(['python', 'manage.py', 'reset', '--no-input'], cb);
 });
@@ -339,6 +336,18 @@ gulp.task('runserver', function(cb) {
     command = command.concat(parameters.filter(String));
 
     spawn('pipenv', command, { stdio: 'inherit' }).on('exit', cb);
+});
+
+gulp.task('generateschema', function(cb) {
+    _runInPipenv([
+        'python',
+        'manage.py',
+        'generateschema',
+        '--title',
+        'Baby Buddy API',
+        '--file',
+        'openapi-schema.yml'
+    ], cb);
 });
 
 /**
--- a/openapi-schema.yml
+++ b/openapi-schema.yml
@@ -1,6 +1,6 @@
 openapi: 3.0.2
 info:
-  title: ''
+  title: Baby Buddy API
   version: ''
 paths:
   /api/children/:


MrApplejuice:
Thanks for running CI. Will work on getting those fixed ASAP :)

api/views.py
```diff
@@ -49,6 +51,40 @@ class TimerViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.TimerSerializer
     filterset_class = filters.TimerFilter
 
+    def __timer_operation(self, pk, func):
+        try:
+            timer = models.Timer.objects.get(pk=pk)
+            return func(timer)
+        except models.Timer.DoesNotExist:
+            return Response(
+                {"detail": "timer does not exist"},
```
cdubz, line 60: We don't use `detail` anywhere else in the API. We do use `non_field_errors` but that doesn't seem to fit here. Lets just return `404` with no body.

api/views.py
```diff
@@ -49,6 +51,40 @@ class TimerViewSet(viewsets.ModelViewSet):
...
+                status=status.HTTP_404_NOT_FOUND
+            )
+
+    @action(detail=True, methods=['patch'])
+    def stop(self, request, pk=None):
+        def do_stop(timer):
+            if not timer.active:
+                return Response(
+                    {"detail": "timer already stopped"},
```
cdubz, line 69: Ditto here re: `detail` -- I'm not actually inclined to indicate the timer is already stopped. I think just returning a successful response would be fine. That said I could also see a `non_fields_errors` response here.

api/views.py
```diff
@@ -49,6 +51,40 @@ class TimerViewSet(viewsets.ModelViewSet):
...
+            timer.stop()
+            return Response({"detail": "timer stopped"})
+        return self.__timer_operation(pk, do_stop)
+
+    @action(detail=True, methods=['patch'])
+    def restart(self, request, pk=None):
+        def do_restart(timer):
+            if timer.active:
+                return Response(
+                    {"detail": "timer already active"},
```
cdubz, line 81: Same note here as L69.

api/views.py
```diff
@@ -49,6 +51,40 @@ class TimerViewSet(viewsets.ModelViewSet):
...
+    @action(detail=True, methods=['patch'])
+    def stop(self, request, pk=None):
+        def do_stop(timer):
+            if not timer.active:
+                return Response(
+                    {"detail": "timer already stopped"},
+                    status=status.HTTP_412_PRECONDITION_FAILED
+                )
+            timer.stop()
+            return Response({"detail": "timer stopped"})
```
cdubz, line 73: In keeping with other `PATCH` operations -- lets return a representation of the updated object here instead.

api/views.py
```diff
@@ -49,6 +51,40 @@ class TimerViewSet(viewsets.ModelViewSet):
...
+    @action(detail=True, methods=['patch'])
+    def restart(self, request, pk=None):
+        def do_restart(timer):
+            if timer.active:
+                return Response(
+                    {"detail": "timer already active"},
+                    status=status.HTTP_412_PRECONDITION_FAILED
+                )
+            timer.restart()
+            return Response({"detail": "timer restarted"})
```
cdubz, line 85: In keeping with other `PATCH` operations -- lets return a representation of the updated object here instead.

cdubz (changes requested):
Thanks, @MrApplejuice! I think these are great enhancements. I just want to tweak the responses a bit to be more in line with the rest of the API.

We'll also want to update the OpenAPI schema doc with these endpoints. There is an automated way to do this that I can't recall off the top of my head...

Hope you're planning to share your app when complete! 😄

api/views.py
MrApplejuice: Ok sounds good. I tried to read the docs for the Django REST framework, and these (including the PATCH-change) seemed to match the behavior from there the best - you can see this when trying to obtain, for example, a timer that does not exist with: http://127.0.0.1:8000/api/timers/1232/

But only a 404 is also fine and logic.

cdubz: Oh wait -- you're right. I didn't realize that. Lets stick with what DRF does for this, and use the same simple detail of `{"detail":"Not found."}`. Thanks for catching.

MrApplejuice: Yeah, no problem. That is why a friendly push and pull can be helpful, right? "Not found" it is!

I will have a look, if I can leverage the code of Django REST Framework somehow for generating this message. Seems like the "nicest" solution but not sure if that is possible somehow. I can namely see on my end that the response messages from the REST API are also localized to German somehow. Probably would be great to just leverage what is there already.

cdubz: Yeah part of why I was thinking to get rid of the detail entirely was one less thing to translate. I'm not sure if DRF does that though... Baby Buddy's language is based on a custom user setting so most certainly not. Regardless, anything we can inherit from DRP is a plus!

api/views.py
MrApplejuice: I was brooding a bit about this, but concur.

If this is not intended, it should be intercepted in the model methods (restart/stop) instead. This might cause possible race conditions that a client cannot prevent: If multiple clients trigger "restart" or "stop" at the same time, start/stop dates will be overwritten.

Not a problem for me, I think pretty much for noone. I will leave it to you to fix this if it really is a problem ;-)

api/views.py
MrApplejuice: (I think this is resolved, I will write a short digest below)

MrApplejuice:
I read up a bit more on how DRF works internally and what is considered "public interface", and I think the new result is way nicer. Quick commented code on where stuff went:

~~~~~~.py
    @action(detail=True, methods=['patch'])
    def restart(self, request, pk=None):
        # This raises a 404 in case pk does not exist, otherwise retrieves 
        # the object - neat
        timer = self.get_object()
    
        # I removed the state-check on the timer - I left a comment above. If intended
        # the state-check can better be implemented in the model's .restart/.stop methods
        timer.restart()

        # Leverage the DRF serializer to render the modified timer-database entry 
        # after a successful patch
        return Response(self.serializer_class(timer).data)
~~~~~~

MrApplejuice:
> We'll also want to update the OpenAPI schema doc with these endpoints.

I found the endpoint `http://127.0.0.1:8000/api/schema` in the code which seems to render that document. 

https://github.com/babybuddy/babybuddy/blob/319ae220fe42e5339ec3615f63a9111472f818f0/api/urls.py#L30-L34

I am not sure if that is how it is supposed to be done however. I could just grab that json file and commit it if that is how it was generated originally?

MrApplejuice:
> Hope you're planning to share your app when complete! smile

That would be the plan - if it is any good. Right now it offers us some utility, but is pretty janky :P

cdubz (approved):
@MrApplejuice this looks great. I tacked a few extra commits for the OpenAPI stuff. Thanks, again, for this contribution!

PR #368, opened 2022-01-10
Fix German translation mistake

MrApplejuice:
"Anzahl" refers to "Count" in German. This always confused me a bit when using the app. "Menge" better refers to "unspecified amount of something". "Menge" sounds weird a bit though, but captures the sentiment of the app the best, which does not specify what unit is used to measure a "diaper change amount". More typical language use would dictate to use something like "Gewicht" (weight) to name what is being measured.

Anyway, this is correct, the previous phrasing was wrong which was naming the graph "Number of diapers".


cdubz:
DDG translate seems to agree with you so I'll take it. Thanks!

dagr1709:
@cdubz I found three more translation errors. The English words are still displayed in three places.

Here are the three, along with what they would be in German.

Add Bottle Feeding - > Flaschenfütterung hinzufügen
Bottle Feeding -> Flaschenfütterung
Breastfeeding -> Stillen

PR #372, opened 2022-01-12
Fix ordering of start/stop items when start and stop times are the same

MrApplejuice:
# Problem

It can be handy sometimes to retroactively add, for example, a "feeding" quickly from the menu without editing the start and end times explicitly. We make use of this feature a lot right when we are done with a feeding but forgot to start a timer for measuring the time it took to do the feeding. This makes it so that start and end times are stored as the same timepoint.

However, when viewing timeline items for these entries, they are ordered the "wrong way" (better said "illogical"), listing "start" after "end":

grafik

_Small nuisance my partner was mentioning, and which was relatively easy to fix now that I am more used to the codebase :-)_

# Solution

Use lexical ordering of `timeline_entry.type` in the template to force the order of entries at equal timepoints:

grafik

### DIFF
--- a/core/templates/timeline/_timeline.html
+++ b/core/templates/timeline/_timeline.html
@@ -17,48 +17,52 @@
 </h3>
 {% if timeline_objects %}
     <ul class="timeline m-auto">
-        {% for object in timeline_objects %}
-            <li{% cycle "" ' class="timeline-inverted"' %}>
-                <div class="timeline-badge {% if object.type == "start" %}bg-success{% elif object.type == "end" %}bg-danger{% else %}bg-info{% endif %}">
-                    <i class="icon-{{ object.model_name }}"></i>
-                </div>
-                <div class="card text-right">
-                    <div class="card-body">
-                        {{ object.event }}
-                        {% for detail in object.details %}
-                            <div><small>{{ detail }}</small></div>
-                        {% endfor %}
+        {% regroup timeline_objects by time as timeline_moments %}
+        
+        {% for moment_objects in timeline_moments %}
+            {% for object in moment_objects.list|dictsort:"type" %}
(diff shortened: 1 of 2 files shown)


core/templates/timeline/_timeline.html
```diff
@@ -17,48 +17,52 @@ <h3 class="text-center">
...
-                <div class="card text-right">
-                    <div class="card-body">
-                        {{ object.event }}
-                        {% for detail in object.details %}
-                            <div><small>{{ detail }}</small></div>
-                        {% endfor %}
+        {% regroup timeline_objects by time as timeline_moments %}
+        
+        {% for moment_objects in timeline_moments %}
+            {% for object in moment_objects.list|dictsort:"type" %}
```
MrApplejuice, line 23: This dictsort does the trick. However, the regroup is necessary to force sorting precedence on time, making this a bit more messy :-/

cdubz (changes requested):

core/templates/timeline/_timeline.html
MrApplejuice: I picked (1)

MrApplejuice:
The change of L27 in timeline.py is a lot more clean. I was not sure if that is "acceptable" in the sense where the line with respect to rendering being agnostic to changes to the data. Anyway. This now is a two-liner and much more expressive with the explicit ordering as a dict.

I reverted the old change as well.

cdubz:
> I was not sure if that is "acceptable" in the sense where the line with respect to rendering being agnostic to changes to the data.

Yeah I tend to struggle with the separation of data and rendering and often get things mixed. So you’d be hard pressed to come up with something I wouldn’t find acceptable in that sense 😁

cdubz (approved):

Issue #374, opened 2022-01-14
Error during add/edit action for fed and diaper


nos86:
Hello, 
I recentrly installed this app in my docker server.

When I try to add or edit an entry for fed or diaper change, I receive this error. If I go back to the dashboard however I find the entry.

The error in app is risen at:
>Exception Value:  unsupported format character '!' (0x21) at index 36
Exception Location: /app/babybuddy/core/views.py, line 37, in get_success_message

The invalid character and its position changes every time, so I assume it is not related with the info I add in the entry.

Please find below the complete traceback

> Environment:
Request Method: POST
Request URL: http://192.168.62.4:8000/changes/add/
Django Version: 3.2.7
Python Version: 3.9.5
Installed Applications:
['api',
 'babybuddy',
 'core',
 'dashboard',
 'reports',
 'axes',
 'django_filters',
 'rest_framework',
 'rest_framework.authtoken',
 'widget_tweaks',
 'easy_thumbnails',
 'storages',
 'import_export',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'babybuddy.middleware.RollingSessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'babybuddy.middleware.UserTimezoneMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'axes.middleware.AxesMiddleware']
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
[...]

PR #375, opened 2022-01-15
add HA Addon link

OttPeterR:
Hey, I love using baby buddy so I put this together. Currently only <your_ip:port> access works -- the ingress doesn't at the moment so maybe we could collaborate and figure that out at some point!

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ or use Baby Buddy in fun ways. If you have a project to share please open a PR
 adding it here or reach out via GitHub Issues or Discussions or on Gitter!
 
 - [Home Assistant integration](https://github.com/jcgoette/baby_buddy_homeassistant)
+- [Home Assistant Addon (host Baby Buddy on your Home Assistant)](https://github.com/OttPeterR/addon-babybuddy)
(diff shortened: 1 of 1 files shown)


cdubz:
Thanks for adding! I'd be happy to try to help with the integration where I can. I was also just looking at #374 with a user that may be related. I'm not at all familiar with HA though. Is there a good guide for getting a test environment up quickly?

OttPeterR:
Here's the guide for addons: https://developers.home-assistant.io/docs/add-ons

From a high level, HA addons are docker containers that open a port for their UI, HA then makes that UI accessible through the HA UI, this lets a bunch of services get funneled through a single URL with added security. The issue I'm having is changing the base URL of baby buddy, because each HA addon needs its own unique url extension. This looks like it maybe be similar to what #186 was having. But the common solution for other addons is to rewrite requests with nginx - and that's where I'm having issues. I know almost nothing about nginx so I'm fumbling around trying to get it to work.

Let's take this discussion off this PR so it can be done with, but I'll sure appreciate any help I can get!

PR #376, opened 2022-01-15
Updated Italian translation

nos86:
File has been generated from poeditor.com after updates of definition


cdubz:
Thanks, @nos86!

Issue #34, opened 2018-03-05
Make dashboard cards configurable per user


cdubz:
Make dashboard cards configurable per user (good example: GitHub's pinned repos system for display and ordering).

[Sortable](https://github.com/RubaXa/Sortable) could work but the current three column dashboard layout makes localStorage save and reload tricky.

cdubz:
Potential for investigation: https://dsmorse.github.io/gridster.js/

qirk4:
This would be a really nice feature! 
I'd like to see the weight data too on the dashboard.

stargalax:
#1086 
this is what I was able to configure locally!

https://github.com/user-attachments/assets/122fe653-095f-4730-8a08-9e488f76280a

currently conducting manual testing to see if things work in different scenarios! just wanted to show the progress I made!

stargalax:
hi! just wanted to give an update , i did manual testing of the feature.. but I have no idea how to add automated tests for the same? so do I raise a PR or should I check for something else before that!
thank you @cdubz

iainsmacleod:
I'd like to see a basic version where I can just choose cards I want to hide, old sleep or pumping data. I can limit data I display to last 4 weeks,  but it's not the same as having configurable cards.

stargalax:
> I'd like to see a basic version where I can just choose cards I want to hide, old sleep or pumping data. I can limit data I display to last 4 weeks, but it's not the same as having configurable cards.

hi! I did implement the hiding feature i,e you can choose the cards you want to hide in the settings! Iwill look into the display of data of a specific time period! @cdubz let me know if that is okay!! thank you

BradNut:
This would be very useful as my partner does not like the recent notes being so prominent on the dashboard. Reordering of dashboard items would also be nice.

PR #382, opened 2022-01-27
Rearrange dashboard cards to set timer card first

adamaze:
Having the timer card show up first is super helpful, especially when you have "Hide Empty Dashboard Cards" set


cdubz:
Thanks for contributing! Curious to see if this gets any negative feedback when it rolls out... maybe an excuse to finally make some time for #34 😄

Issue #383, opened 2022-01-30
Show type breakdown on feeding_amounts report


alzyee:
I would like to see the breakdown of various feeding types in the feeding amount report. I don't do much python coding so my code is just a hack to get the thing I wanted as a demo. Ideally it would show only / all the types that have values.

https://github.com/alzyee/babybuddy/commit/34b6ecaf22f84b833cf4a36ebd2aada7e6253fd6

cdubz:
Sounds interesting and the code looks like a good start, if you open a draft PR I can take a closer look in review.

alzyee:
I think I see a route to implement the features I suggested. It seems like I can make an array of dictionaries then into an array of traces which can be passed into the  data attribute of the figure. I will make an effort to get the code functioning as described and make a PR.

* show only / all the types that have values
* breakdown of various feeding types

cdubz:
Closed by #384.

PR #384, opened 2022-01-30
breakout feeding types

alzyee:
There are some bugs that this change appears to create.
* on mobile it puts the legend over the graph
* on mobile it makes the trend very narrow 

I tried adding this code to the layout but on mobile it overlaps the x axis labels
        `'legend':{
            'orientation':'h'
        }`

It might be better to just not have a label. 
in reference to #383

### DIFF
--- a/reports/graphs/feeding_amounts.py
+++ b/reports/graphs/feeding_amounts.py
@@ -7,30 +7,63 @@ import plotly.graph_objs as go
 
 from reports import utils
 
-
 def feeding_amounts(instances):
     """
     Create a graph showing daily feeding amounts over time.
     :param instances: a QuerySet of Feeding instances.
     :returns: a tuple of the the graph's html and javascript.
     """
-    totals = {}
+    feeding_types=[
+            'breast milk',
+            'formula',
+            'fortified breast milk',
+            'solid food',
+        ]
+    feeding_types_desc=[
+            'Breast milk',
+            'Formula',
+            'Fortified breast milk',
+            'Solid food',
+        ]
+    total_idx=len(feeding_types)+1 #+1 for aggregate total
+    totals_list=list()
+    for i in range(total_idx):
+        totals_list.append({})
     for instance in instances:
         end = timezone.localtime(instance.end)
         date = end.date()
-        if date not in totals.keys():
-            totals[date] = 0
-        totals[date] += instance.amount or 0
+        if date not in totals_list[total_idx-1].keys():
+            for item in totals_list:
+                item[date]=0
+        feeding_idx=feeding_types.index(instance.type)
+        totals_list[feeding_idx][date]+= instance.amount or 0
+        totals_list[total_idx-1][date]+= instance.amount or 0
+    zeros = [0 for a in totals_list[total_idx-1].values()]
+    
+    #sum each feeding type for graph
+    amounts_array=[]
+    for i in range(total_idx):
+        amounts_array.append([round(amount, 2) for amount in totals_list[i].values()])
+    
+    traces=[]
+    for i in range(total_idx-1):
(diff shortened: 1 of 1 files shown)


reports/graphs/feeding_amounts.py
```diff
@@ -14,23 +14,57 @@ def feeding_amounts(instances):
     :param instances: a QuerySet of Feeding instances.
     :returns: a tuple of the the graph's html and javascript.
     """
-    totals = {}
+    feeding_types = [
```
cdubz, line 17: You can use model metadata to get these types directly from the model (after its imported). Some variation of `feeding_types = Feeding._meta.get_field('type').choices` should do the trick.

reports/graphs/feeding_amounts.py
```diff
@@ -14,23 +14,57 @@ def feeding_amounts(instances):
...
+            'breast milk',
+            'formula',
+            'fortified breast milk',
+            'solid food',
+        ]
+    feeding_types_desc = [
```
cdubz, line 23: See previous comment re: model field metadata. These descriptive names will come from the same place.

reports/graphs/feeding_amounts.py
```diff
@@ -14,23 +14,57 @@ def feeding_amounts(instances):
...
+    for i in range(total_idx):
+        amounts_array.append([round(a, 2) for a in totals_list[i].values()])
+
+    traces = []
+    for i in range(total_idx-1):
+        traces.append(go.Bar(
+            name=feeding_types_desc[i],
+            x=list(totals_list[total_idx-1].keys()),
+            y=amounts_array[i],
+            text=amounts_array[i]
```
cdubz, line 55: On hover right now we get too much information -- date, amount again, and type. Ideally we just have type.

image

I'm not 100% sure how to set it up that way -- take a look at https://plotly.com/python/hover-text-and-formatting/ and if ya get stuck I can take a closer look.

reports/graphs/feeding_amounts.py
```diff
@@ -14,23 +14,57 @@ def feeding_amounts(instances):
...
+
```
cdubz, line 56: Remove this empty space.

reports/graphs/feeding_amounts.py
```diff
@@ -14,23 +14,57 @@ def feeding_amounts(instances):
...
         hoverinfo='text',
         textposition='outside',
-        text=amounts
-    )
+        text=amounts_array[total_idx-1],
+        showlegend=False
+    ))
 
     layout_args = utils.default_graph_layout_options()
     layout_args['title'] = _('<b>Total Feeding Amounts</b>')
```
cdubz, line 70: Lets update the title of this graph to "Total Feeding Amount by Type". We'll need to do this elsewhere as well.

cdubz (changes requested):
This a great start! A few change requests here. Let me know if you need any help with these changes.

reports/graphs/feeding_amounts.py
alzyee: I changed as suggested but it seems like this should somehow come from the translations. I'm not sure how to make that happen.

alzyee:
I think I addressed all of your comments

reports/graphs/feeding_amounts.py
cdubz: @alzyee yep no worries -- that is what `_` does. It's a weird Python convention: https://github.com/babybuddy/babybuddy/blob/master/reports/graphs/feeding_amounts.py#L3. Translations will need to be updated to pick up the change but I'll do that separately.

cdubz (approved):
Looks great! Thanks, @alzyee.

PR #386, opened 2022-02-06
Hide feeding_day card when no information present

BenjaminHae:
When no amount is stored with feedings, this card does not show information, so it can be hidden.


cdubz:
It does show number of feedings for the day even if amounts are not used but maybe that information isn’t really worth much. I’ll take this as-is and continue to look forward to the day when I can actually take the time to make the dashboard cards more configurable (:

PR #396, opened 2022-02-20
Tagging

### LINKED ISSUE #380: Feature suggestion: Add user editable tags to "notes"-sections
# Problem

We find ourselves using the "notes" function quite a bit for writing taking medical notes or other standard findings that we would like to keep track of. These tend to be single-events like "Took a bath" or similar, so very much fitting into the "notes" category. However, for some of them (medical ones especially), they tend to become _very_ repetitive and cumbersome to type out: "Took vitamins" ... every day... and then often on a phone...

# Suggested solution

Pretty much all of the different pages have the handy "notes" field already:

grafik

Instead of typing out the note every time, I propose adding another "tags" section (like the labels here on the issues page), that allows quickly adding a tag to the entry - just like a note. But instead of typing, these will be stored in the database and are quickly accessible from the interface with a simple click. See doodle below:

grafik

(I just realized that adding _these_ tags to a "temperature" entry does not make a lot of sense - picture those tags on the generic "notes" page instead, maybe)

The tags would be stored in a new model in the database and the most recent 10(?) would be offered as a quick-list (most recently used by user, maybe?). Alternatively the (+) can be clicked to create a new tag on-the-fly. Text and color are user-editable. The design would be the "bootstrap pill" or "bootstrap badge" style (https://getbootstrap.com/docs/5.0/components/badge/).

For the database design, I propose adding a new table "tags" to the core-models and then adding a Many-To-Many relation for all items that can contain "Notes" right now to that table.

# Other related issues

I have found #217 which could also be a solution to this issue (create a custom event "took vitamins", "took baby bath"). However, to me it looks like that issue is way more complex to solve migrations and proper database design. This would solve the "single repetitive events" usecase neatly and quickly. In addition, I think that this might also offer utility once #217 is implemented as this feature could be recycled as a means to tag (and therefore index) events across different "types". Seems handy to me, in cases where someone wants to tag "will be done by dad/mom" or something. Point is: Even if #217 is implemented, I think that this offers utility.

-------------------

I would be willing to give this a go as I was thinking of shoehorning this into my Android-app, but then was thinking that proper server support would be even nicer. Before working (wasting) tome on the feature some green-lighting would be nice :-)

MrApplejuice:
Fixes #380

This took quite a bit longer than anticipated. Sorry for that. I had some problems with integrating taggit into BabyBuddy (see code comment). ~**This is a "preview" of the new feature**, it is rather trivial to complete from here: Code cleanups, adding tags to other models, maybe add tags to a few other views.~ Update: This is PR is ready.

# New features/changes

- Added the taggit library for support (thanks for the tip)
- Added custom tag-model with 4 fields: name, slug, color, and "last_used"
    - name + slug are taggit defaults
    - the `color` field for assigning unique colors to tags
    - the `last_used` field is for ordering tags in a recently-used list in the dedicated widget (see below)
- Added a dedicated widget for editing tags:
grafik
    - Allows to quickly and conveniently add tags from a "recently used tags" list
- Added a REST-endpoint to list and create notes under `http://localhost:8000/api/tags/`
    - Supported filters: `name` and `last_used`
    - Example:
~~~~~~
{
    "count": 5,
    "next": null,
    "previous": null,
    "results": [
        {
            "slug": "some-tag",
            "name": "Some TAG",
            "color": "#007F00",
            "last_used": "2022-02-20T11:43:19.710098Z"
        },
[...]
~~~~~~

I recorded a demo-video of how this all looks and works:

https://user-images.githubusercontent.com/542105/154842589-949f098e-4a22-413b-81c6-9acc445fab51.mp4

# Todo

Before this is merged the following problems should be addressed, IMO:

- [x] Linting does not pass
- [x] JavaScript needs to be removed from the widget-html code
- ~Tags should be added to most items of the database (not only "notes")~
    - I will do this in a separate PR - this one is big enough
- [x] Tests for this new, fresh feature

I made this a draft to allow you to verify that this is looking ok/as intended and generally going in the right direction :-)


core/forms.py
```diff
@@ -82,6 +82,7 @@ def save(self, commit=True):
             timer.stop(instance.end)
         if commit:
             instance.save()
+            self.save_m2m()
```
MrApplejuice, line 85: :exclamation:  This one cost me a week or so. Without this line many2many relations are not being updated. It might be the case that this was a bug in BabyBuddy in general? Who knows, this makes it all work :-)

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,233 @@
...
+        </span>
+        {% endfor %}
+    </div>
+    <input 
+        type="hidden"
+        name="tags"
+        value="{% for t in widget.value %}&quot;{{ t.name }}&quot;{% if not forloop.last %},{% endif %}{% endfor %}"
+    >
+<script>
+        window.addEventListener('load', () => {
```
MrApplejuice, line 39: This should be in its own `*.js` file. First version is inlined for easy of prototyping.

core/widgets.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Dict, Optional
+from django.forms import Widget
+
+class TagsEditor(Widget):
```
MrApplejuice, line 4: Custom widget to allow editing tags

core/models.py
```diff
@@ -70,6 +81,45 @@ def validate_time(time, field_name):
...
+TAG_COLORS = [
+    "#FF0000", "#00FF00", "#0000FF", "#FF00FF", "#FFFF00", "#00FFFF",
+    "#FF7F7F", "#7FFF7F", "#7F7FFF", "#FF7FFF", "#FFFF7F", "#7FFFFF",
+    "#7F0000", "#007F00", "#00007F", "#7F007F", "#7F7F00", "#007F7F",
+]
+
+def random_color():
+    return TAG_COLORS[random.randrange(0, len(TAG_COLORS))]
+
+class BabyBuddyTag(TagBase):
```
cdubz, line 93: Is there a reason to prefix this with `BabyBuddy` instead of just calling the model `Tag` (would be my preference)?

core/models.py
```diff
@@ -70,6 +81,45 @@ def validate_time(time, field_name):
...
         )
 
+TAG_COLORS = [
+    "#FF0000", "#00FF00", "#0000FF", "#FF00FF", "#FFFF00", "#00FFFF",
+    "#FF7F7F", "#7FFF7F", "#7F7FFF", "#FF7FFF", "#FFFF7F", "#7FFFFF",
+    "#7F0000", "#007F00", "#00007F", "#7F007F", "#7F7F00", "#007F7F",
+]
+
+def random_color():
+    return TAG_COLORS[random.randrange(0, len(TAG_COLORS))]
```
cdubz, line 91: I don't think we need `TAG_COLORS` separated from `random_color()`. Ideally we could be a bit smarter here too (e.g. generating colors with good contrast and not reusing colors) but we can save that for a future iteration.

core/models.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
+import re
```
cdubz, line 2: Is this import used anywhere?

core/models.py
```diff
@@ -9,7 +12,15 @@
...
 from django.utils.translation import gettext_lazy as _
+from django.utils.timezone import now
+from django.core.validators import RegexValidator
 
+import random
+
+from taggit.managers import TaggableManager
+from taggit.models import TagBase, GenericTaggedItemBase, TaggedItemBase
+
+random.seed(time.time())
```
cdubz, line 23: Instead of the extra import lets just use `now().timestamp()` to seed.

cdubz: Also it is necessary to seed here? Any reason _not_ to see in `random_color` specifically?

core/forms.py
cdubz: Interesting find! This might be the first m2m relationship Baby Buddy has so it makes sense this hasn't come up before...

Reading through [Creating forms from models > the `save()` method](https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/#the-save-method) it sounds like this should be done anywhere `commit=False` is used. This is the case in a few other places in the code.

core/models.py
```diff
@@ -9,7 +12,15 @@
 from django.utils import timezone
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
+from django.utils.timezone import now
```
cdubz, line 15: Note that `timezone` is already imported above -- lets ditch this extra import.

core/models.py
```diff
@@ -70,6 +81,45 @@ def validate_time(time, field_name):
...
+    
+    color = models.CharField(
+        "Color",
+        max_length=32,
+        default=random_color,
+        validators=[RegexValidator(r"^#[0-9A-F]{6}$")]
+    )
+
+    last_used = models.DateTimeField(
+        default=now,
```
cdubz, line 106: Use `timezone.now` as the default here.

core/models.py
```diff
@@ -70,6 +81,45 @@ def validate_time(time, field_name):
...
+
+class BabyBuddyTagged(GenericTaggedItemBase):
+    tag = models.ForeignKey(
+        BabyBuddyTag,
+        on_delete=models.CASCADE,
+        related_name="%(app_label)s_%(class)s_items",
+    )
+
+    def save_base(self, *args, **kwargs):
+        self.tag.last_used = now()
```
cdubz, line 119: Use `timezone.now()` here as well.

core/models.py
```diff
@@ -70,6 +81,45 @@ def validate_time(time, field_name):
...
+        validators=[RegexValidator(r"^#[0-9A-F]{6}$")]
+    )
+
+    last_used = models.DateTimeField(
+        default=now,
+        blank=False,
+    )
+
+
+class BabyBuddyTagged(GenericTaggedItemBase):
```
cdubz, line 111: Ditto here re: the prefix -- any reason we can't just use `Tagged` for the model name?

core/widgets.py
```diff
@@ -0,0 +1,44 @@
...
+    input_type = 'hidden'
+    template_name = 'core/widget_tag_editor.html'
+
+    @staticmethod
+    def __unpack_tag(tag):
+        return {'name': tag.name, 'color': tag.color}
+
+    def format_value(self, value: Any) -> Optional[str]:
```
cdubz, line 12: Just a comment -- this is the first time I have seen typing used. I'm definitely a big fan of this other languages and would like to refactor Baby Buddy to use it eventually.

core/widgets.py
```diff
@@ -0,0 +1,44 @@
...
+        if value is not None and not isinstance(value, str):
+            value = [self.__unpack_tag(tag) for tag in value]
+        return value
+    
+    def get_context(self, name: str, value: Any, attrs) -> Dict[str, Any]:
+        from . import models
```
cdubz, line 18: Any reason not to import this at the top? Just not a pattern I have used myself.

core/widgets.py
```diff
@@ -0,0 +1,44 @@
...
+
+        most_tags = models.BabyBuddyTag.objects.order_by(
```
cdubz, line 20: Why only "most"?

core/widgets.py
```diff
@@ -0,0 +1,44 @@
...
+        result['widget']['tag_suggestions'] = {
+            'quick': [
+                self.__unpack_tag(t) for t in quick_suggestion_tags
+                if t.name not in tag_names
+            ],
+            'most': [
+                self.__unpack_tag(t) for t in most_tags
+            ]
+        }
+        return result
```
cdubz, line 44: Generally I'd like to see more inline documentation in this file explaining what is being done and why.

package-lock.json
```diff
@@ -1,9921 +1,7 @@
 {
```
cdubz: What happened here? I don't think we want to remove this lock data 😄

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,233 @@
...
+            function createTagClicked() {
+                const tagName = addTagInput.value.trim();
+                const uriTagName = encodeURIComponent(tagName);
+
+                function success() {
+                    addTagInput.value = "";
+                }
+
+                function fail(msg) {
+                    msg = msg || "Error creating tag";
```
cdubz, line 86: This will need to be translatable. I have never looked in to string translation in JS with Django so not sure how challenging that will be. Perhaps we can just refactor this string in to the twig template.

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,233 @@
...
+
+                    addTagInput.select()
+                    alert(msg);
+                }
+
+                if (!tagName) {
+                    fail('Not a valid tag name');
```
cdubz, line 93: This will need to be translatable. I have never looked in to string translation in JS with Django so not sure how challenging that will be. Perhaps we can just refactor this string in to the twig template.

core/templates/core/widget_tag_editor.html
cdubz: I'll save some review for once we get this a little further along (I'm also just in Python brain right now and don't feel like switching to JS 😃)

cdubz (changes requested):
@MrApplejuice this is a great start -- thank you! I think you have the right stuff in mind for moving this forward. Once you get the `requirements.txt` updated (see review comments) this will be deployable to Heroku so we can poke around in the environment there.

Also re: linting that should be much easier now since we have adopted Black -- just run `gulp format` before commits. I forgot to add note of this new command to the docs 😬

core/templates/core/widget_tag_editor.html
MrApplejuice: I think it would be ideal to put this in a statically configured bootstrap-modal, for example like this: https://getbootstrap.com/docs/4.4/components/modal/#live-demo

Then everything can stay in the template and we can translate using template features only. Also we do not need to use the awkward out-of-place looking alert function.

core/templates/core/widget_tag_editor.html
MrApplejuice: Yes, please do not review the js file yet! That thing is yucky right now... very much prototype-stuff. I wanted to show off how everything looks & feels with this draft and what changes to the python-code are necessary. Frontend-stuff typically is less likely to leave traces in the codebase for the long-haul imo, so I did not put much effort in the js-stuff yet.

core/forms.py
MrApplejuice: I think this will be interesting. I found this in the root form class and thought it covers most/all forms. I will try to find more instances, but not sure if I manage to find them all...

MrApplejuice:
Hi, I am working on the JavaScript code, doing some cleanup etc and was wondering: Is there a coding style/conventions babybuddy follows? I looked a bit in files like babybuddy.js and am mainly wondering about used standards:

I see that jQuery is used prominently but not so much more modern constructs like new-style lambda functions `() => {}`, for loops `for (const i of [1, 2, 3])`, or async functions - all of which I would find handy to use :-) 

So I was wondering - is it ok for me to use those newer language features? Or should I rewrite my code to "good ol' jQuery"? Any ES-version I should target?

Sidenote: I still like jQuery, but the new language features seem superior to me.

cdubz:
@MrApplejuice no standards to speak of! My JS skills were very limited back when I started work on Baby Buddy and haven't improved _much_ since then 😄 Happy to avoid jQuery and stick to newer language features _but_ I do want to maintain somewhat flexible support for older browsers. I would be more than happy if you want to take the initiative to establish some form of standard and maybe at least some handling of minimum supported browsers?

MrApplejuice:
Hey, thanks for the update.

I think it is best to recode my stuff with more tranditional javascript features then. I am certainly not a JavaScript-Guru myself, but we have one at my nine-to-five and from what I understand from him the typical way to guaratee backwards compatibility is to use webpack or sth to build backwards compatible bundles from "modern javascript". That way you get the best of two worlds: backwards compatibility plus you can use all "them newfangled features" of the newer javascript versions...

Anyway, I think this goes beyond what I am trying to achieve in this PR. Vanilla jQuery it is!

------

After writing this, I think it would still be handy to specify what "vanilla" means. ES6 would be nice to target and has a lot of "nice language features" . ES5 would be the older alternative. Here a quick overview of wrt browser compatibility date etc:

- ES5 (browser support 98%, since ~2013) - Sources: [caniuse](https://caniuse.com/?search=ES5) | [syntax overview](https://www.w3schools.com/Js/js_es5.asp)
- ES6 (browser support 95%; since ~2017) - Sources: [caniuse](https://caniuse.com/?search=ES6) | [syntax overview](https://www.w3schools.com/Js/js_es6.asp)

Would ES6 be an acceptable choice until someone incorporates webpack for compatibility transpiling?

cdubz:
@MrApplejuice yep ES6 sounds good!

babybuddy/static_src/scss/global.scss
```diff
@@ -52,3 +52,8 @@
 .icon-2x {
     font-size: 1.65em;
 }
+
+// All modals
+.modal-content {
+    color: theme-color('dark');
+}
```
MrApplejuice, line 59: Without this, the modal was still rendered in the "light" theme and was barely readable.

babybuddy/templates/babybuddy/form.html
```diff
@@ -1,9 +1,12 @@
 {% load i18n widget_tweaks %}
 
+{# Load any form-javascript files #}
+{{ form.media.js }}
```
MrApplejuice, line 4: Important generic addition :eyes:

core/models.py
MrApplejuice: Actually, it was just "muscle memory" - probaby from C. Default seed is ok as well, I think :)

core/widgets.py
MrApplejuice: VS Code is inserting them manually for me. And in my 9-5, we are using them pretty much everywhere. Very handy, specially when combining with an IDE that understands the type annotations so you get your handy Ctrl+Space auto completion! `Any` is a bit of a disappointing annotation here correct in this case, from what I was able to derive.

Anyway, I would add them as-you-go to the project. Slowly they will build up and become helpful. IMO. Anyway, I hope you do not mind that I add them?

core/widgets.py
MrApplejuice: Nope... just another case of muscle memory. Only needed for admin-panels I think(?) Whatever, not applicable in this case :D

cdubz:
@MrApplejuice one random thing I forgot -- Heroku doesn't like pipenv so when you make changes to installed packages you also need to do `pipenv lock -r > requirements.txt` and commit the changes to `requirements.txt`. This will get Heroku running for the branch now that linting issues are resolved 👍

babybuddy/static_src/scss/global.scss
cdubz: Interesting. I'll take a look at this when I do some testing. May want to customize further.

core/widgets.py
cdubz: No problem -- add away! I'll probably start adding them myself after this gets merged in and I get the hang of it.

core/models.py
MrApplejuice: Removed the prefix now

package-lock.json
MrApplejuice: Some sort of accident... I actually do not know o_O - Reverted

MrApplejuice:
> @MrApplejuice one random thing I forgot -- Heroku doesn't like pipenv so when you make changes to installed packages you also need to do `pipenv lock -r > requirements.txt` and commit the changes to `requirements.txt`. This will get Heroku running for the branch now that linting issues are resolved +1
> 
> _Edit: Oops I just noticed I did [already reference this](https://github.com/babybuddy/babybuddy/pull/396#discussion_r810761369) smiley. [It's in the docs now, too!](https://docs.baby-buddy.net/contributing/pull-requests/#pipfile)_

Yeah, I remembered, just did not have time to work on this - Cannot work every evening on this, afterall there is a baby to take care for here as well :P

I think most of your comments are handled now. I will fix the conflicts now, and then add some tests. Then code quality should be there.

core/models.py
```diff
@@ -71,6 +80,71 @@ def validate_time(time, field_name):
...
+    ]
+    return TAG_COLORS[random.randrange(0, len(TAG_COLORS))]
+
+
+class Tag(TagBase):
+    class Meta:
+        verbose_name = _("Tag")
+        verbose_name_plural = _("Tags")
+
+    color = models.CharField(
```
cdubz, line 112: Add a translated `verbose_name` to this field.

core/models.py
```diff
@@ -71,6 +80,71 @@ def validate_time(time, field_name):
...
+        "Color",
+        max_length=32,
+        default=random_color,
+        validators=[RegexValidator(r"^#[0-9A-F]{6}$")],
+    )
+
+    last_used = models.DateTimeField(
```
cdubz, line 119: Add a translated `verbose_name` to this field.

core/models.py
```diff
@@ -71,6 +80,71 @@ def validate_time(time, field_name):
...
+        default=timezone.now,
+        blank=False,
+    )
+
+
+class Tagged(GenericTaggedItemBase):
+    tag = models.ForeignKey(
```
cdubz, line 126: Add a translated `verbose_name` to this field.

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,64 @@
...
+        {{ k }}="{{ v }}"
+    {% endfor %}>
+    {% csrf_token %}
+    <span class="prototype-tag btn badge badge-pill cursor-pointer mr-1" style="display: none;">
+        UNINITIALIZED PROTOTYPE
+        <span class="add-remove-icon pl-1 pr-1">+ or -</span>
+    </span>
+    <div class="current_tags" style="min-height: 2em;">
+        {% for t in widget.value %}
+        <span data-value="{{ t.name }}" data-color="{{ t.color }}" class="tag btn badge badge-pill cursor-pointer mr-1" style="background-color: {{ t.color }};">
```
cdubz, line 14: Lets add a simple `title` attribute to these tags, "Remove tag".

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,64 @@
...
+    <div class="new-tags">
+        <div class="create-tag-inputs input-group">
+            <input class="form-control" type="text" name="" placeholder="{% trans "Tag name" %}">
+            <div class="input-group-append">
+                <button class="btn btn-outline-primary bg-dark btn-add-new-tag" type="button">{% trans "Add" %}</button>
+            </div>
+        </div>
+        <span>{% trans "Recently used:" %}</span>
+        {% for t in widget.tag_suggestions.quick %}
+        <span data-value="{{ t.name }}" data-color="{{ t.color }}" class="tag btn badge badge-pill cursor-pointer mr-1" style="background-color: {{ t.color }};">
```
cdubz, line 29: Lets add a simple `title` attribute to these tags, "Add tag".

core/admin.py
```diff
@@ -177,3 +177,16 @@ class WeightAdmin(ImportExportMixin, ExportActionMixin, admin.ModelAdmin):
...
     )
     resource_class = WeightImportExportResource
+
+
+class TaggedItemInline(admin.StackedInline):
+    model = models.Tagged
+
+
+@admin.register(models.Tag)
+class TagAdmin(admin.ModelAdmin):
```
cdubz, line 187: Although I don't think we do this anywhere else... it would nice to override the admin form and set the color field `type` to `color`. See e.g. https://stackoverflow.com/a/39864918

babybuddy/settings/base.py
```diff
@@ -26,6 +26,7 @@
     "api",
     "babybuddy",
     "core",
+    "taggit",
```
cdubz, line 29: I wonder -- can we remove this to prevent adding the default model (i.e., what shows up in the Django admin interface under "Taggit") since we have out own custom model?

Unclear if that is what is meant by this note --

> Note: Including ‘taggit’ in settings.py INSTALLED_APPS list will create the default django-taggit and “through model” models. If you would like to use your own models, you will need to remove ‘taggit’ from settings.py’s INSTALLED_APPS list.

From https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#using-a-custom-tag-or-through-model

core/templates/core/widget_tag_editor.html
```diff
@@ -0,0 +1,64 @@
...
+        </span>
+        {% endfor %}
+    </div>
+    <input 
+        type="hidden"
+        name="{{ widget.name }}"
+        value="{% for t in widget.value %}&quot;{{ t.name }}&quot;{% if not forloop.last %},{% endif %}{% endfor %}"
+    >
+
+    <div class="modal fade tag-editor-error-modal">
```
cdubz, line 41: Definitely want to do some more custom styling on this but we could always save that for a follow-up issue.

babybuddy/static_src/js/tags_editor.js
```diff
@@ -0,0 +1,337 @@
...
+        return parseInt(x, 16);
+    }
+
+    /**
+     * Attempt to compute a high-contrast color from a background color.
+     * 
+     * (This probably should be researched better because this was 
+     * hand-crafted ad-hoc.)
+     */
+    function computeComplementaryColor(colorStr) {
```
cdubz, line 15: Perhaps we can just pull from this MIT licensed snippet? https://vanillajstoolkit.com/helpers/getcontrast/

babybuddy/static_src/js/tags_editor.js
```diff
@@ -0,0 +1,337 @@
...
+            list.appendChild(tag);
+            this.callTagListUpdatedListeners();
+        }
+    };
+
+    /**
+     * Handler for the edit field allowing to dynamically create new tags.
+     * 
+     * Handles user inputs for the editor. Calls the 'onInsertNewTag' callback
+     * when the craetion of a new tag has been requested. All backend handling
```
cdubz, line 138: Misspelling of "creation" here.

cdubz (changes requested):
@MrApplejuice did another round or review and tested in the PR heroku env -- this is looking really great!

babybuddy/static_src/js/tags_editor.js
MrApplejuice: Excellent idea. At some point it might be worth it to include a npm-module for this kind of stuff, but this works well! :+1:

MrApplejuice:
> @MrApplejuice did another round or review and tested in the PR heroku env -- this is looking really great!

Thank you! Nice to hear that it works to your liking :smiley:

core/widgets.py
MrApplejuice: Forgot to comment on this. "most" to prevent DDoS attacks in case someone adds thousands of tags. That will make the database operation very expensive, so I cap it at 256. Just like REST-API caps requests on pages at 100 requests. This however is a direct database operation, so I had to do it myself.

core/admin.py
MrApplejuice: I did not even know that this is a thing. I will have to read up on it first... sounds like a good idea nonetheless!

babybuddy/settings/base.py
MrApplejuice: Ah yes! That seems to work. I had problems removing the app before, but cannot recall what that was. Now, with all custom models it seems to work. Could not spot any issues. Good one! :+1:

core/templates/core/widget_tag_editor.html
MrApplejuice: I think the title-attribute is the last open issue... I was thinking a bit about it and still wonder a bit what the idea behind it would be: I assume it is to disambiguate between the internal identifier and the shown title, right? If so, shouldn't we choose slug and name instead? Slug is not really used in the frontend however, because I discovered that taggit refers to tags by "name" by default, so they are pretty much a primary-key as well. That is why there is only a "name" at the moment.

I can am not opposed to adding still add `data-title` as well but it would just be equal to `data-name`. Is that acceptable?

cdubz: Really all I’m thinking is there should be some kind of indicator about what clicking on the tags does. It’s OK as-is but the title might just be a helpful addition. I’m not totally attached to it.

MrApplejuice:
Hey... the only open issue is the "data-title" thing for this PR.

I decided to add tags to more forms and models in a separate PR. I think this one is complex enough already and the remaining changes for adding tags to the other forms and models is more a kind of a "big boilerplate operation": Lots of changes but all pretty simple. This PR is a feature-PR and rather self-contained in that.

I will open the other PR against this one.

cdubz:
Makes sense! I would also like to think about how to filter by tags in the forms but that can be another PR as well think.

MrApplejuice:
> Makes sense! I would also like to think about how to filter by tags in the forms but that can be another PR as well think.

Yes, that would be really handy, too. I expect that this will go hand-in-hand with some tougher UI choices that I am not willing to commit to yet ;-)

In addition there are a few not-so-tough additions that I propose to make this really handy:
- I would like to add the tags to the timeline view and the list views of the different catergories
- I laid the groundwork for auto-completion support for the text-field of the tags-editor, but did not actually implement it (keyword: "most_tags")
- Having some interface other than the admin-page to edit tags and tag-colors would be really nice.

core/templates/core/widget_tag_editor.html
MrApplejuice: Ah! So you mean an actually visible ui-element, right?

cdubz:
Yep. All makes sense. I was thinking about mentioning a custom admin UI but as you say this PR is beefy enough already.

core/templates/core/widget_tag_editor.html
cdubz: Yeah. `title` is just a cheap/easy way to do that.

MrApplejuice: I substituted the title-idea with the help_text, mainly because I could not get the "title" idea to "look right".

grafik

Could probably be styled better though, but I think it does the trick.

core/admin.py
MrApplejuice: Done! :) There now is a nice color-picker in the admin UI.

MrApplejuice:
I _think_ that is it... the tags-editor for "Notes" is working and this PR is finished.

MrApplejuice:
Alright.... because it is daytime (here) and I got some time, I prepared a few quick-fire follow-up PRs. Need to wait for this one to be merged first, however. Sneak peek:

image

image

cdubz (approved):
Let’s do it!

MrApplejuice:
> Merged! Thanks, @MrApplejuice! I did notice one thing but didn’t think it worth blocking — the plural name of the model in the Django admin is “Tagss”. We can fix that in another PR.

Darn it! I had actually changed this to make a api-test work, but did not think much of it.

Check this commit: https://github.com/MrApplejuice/babybuddy/commit/7d8ead594c28f6d3479b3cd0580e7d67ea2d3a88

Problem was that the API-tests contain a test for the list-names, and that got very weird with the plural form. I do not remember the details but the test failed. Maybe we need to reintroduce the plural form and just add an exception for the test or sth...

PR #404, opened 2022-02-22
Update Spanish translations

Alberdi:
Imported from POEditor. Again, I don't have any clue why it messes so bad with the files.


cdubz:
Yeah, I don't know either but ultimately the changes are fine and POEditor makes the translation more accessible so I'll take it.

Thanks, again, for contributing!

PR #405, opened 2022-02-23
All the words have been translated into Chinese.

gesisirus:
Chinese can be displayed correctly after renaming folder zh_hans to folder zh


cdubz:
Interesting. Django uses `zh_Hans` and that worked in my local environment but did seem to have issues in other environments. Changing to `zh` does appear to be better. Thanks for the translations _and_ for the fix!

PR #414, opened 2022-03-08
Add tags support to most things

MrApplejuice:
Except: Timers and Children

Details:
- Adds the tags widget to a bunch of models
- Makes the TagsEditor the default editor for all TagsManager-fields
    - This allows us to remove the `widget` field from all forms that contain tags

(No real a screenshot for this one - it is the feature from #396 but now available for weights/BMI/feedings)


cdubz (approved):
👍

PR #416, opened 2022-03-08
Show tags in lists

MrApplejuice:
Addition for #396 

This contains the changes resulting in tags showing up in lists and the timeline:

image
image

Detailed changes:
- Adapted a bunch of templates
    - _Sidenote: I understand why #217 is a thing. Lots of duplication in those templates. So much work to adapt them :P_
- I had to reimplement complenetary_color for the Tag-model because otherwise it cannot be used in Python. So it is there twice now - once in javascript and once in Python.


cdubz (approved):

cdubz:
Haha yeah there is lots of boilerplate 😬

PR #422, opened 2022-03-26
Update german translations

myxor:
Update german translations.

I just exported the files from POEditor - not sure if this is correct.


cdubz:
Thanks, @myxor! POEditor exports always seem to mangle the files a bit but not in any destructive way. This looks good 👍

PR #431, opened 2022-04-06
Allow diaper change without any contents

### LINKED ISSUE #426: Diaper change without wet/solid
For me it makes sense to add a diaper change event without selecting any of _wet_ and _solid_ types.

We sometimes change the diaper for the child to feel more fresh even though nothing is actually in it.

I would like to still keep track of these changes but currently it is not allowed to add such an event without selecting wet and solid state.

myxor:
Fixes #426.

It looks like the model change is enough for this change.

image

### DIFF
--- a/babybuddy/management/commands/fake.py
+++ b/babybuddy/management/commands/fake.py
@@ -159,8 +159,6 @@ class Command(BaseCommand):
         color = ""
         if solid:
             color = choice(models.DiaperChange._meta.get_field("color").choices)[0]
-        if not wet and not solid:
-            wet = True
         amount = Decimal("%d.%d" % (randint(0, 6), randint(1, 9)))
         time = self.time + timedelta(minutes=randint(1, 60))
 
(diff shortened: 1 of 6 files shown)


cdubz:
This also affects the[`card_diaperchange_types` card](https://github.com/babybuddy/babybuddy/blob/283e30cdaace54678c00dd4937ca20b4392fbbdd/dashboard/templatetags/cards.py#L52) but I'm not totally sure how best to address this change there. The idea of that card is to give a quick useful view of how often the child pee'd and pooped for the past few days (i.e., it's not necessarily about the _changes_ themselves).

Just the total number of changes generally is not really as useful and if a user did not use wet or solid the card would just be empty. But maybe that's OK?

myxor:
Maybe that's okay.
Another idea is that the card could contain a third bar called "empty" or "none" or something to represent the complete number of changes?

cdubz:
Yeah but the "empty"/"none" group gets a little dicey because its also possible for a change to be wet _and_ solid so you could have e.g. three diapers -- two wet and solid and one with none -- and the bar could end up look like 2 wet, 2 solid, 1 none. I guess that is fine though...

myxor:
Would be fine for me.

cdubz:
Merged! Thanks!

PR #435, opened 2022-04-10
Dashboard api

rubenixnagios:
How can i add a translation? I already added onto weblate the catalan translation
will it be available on next release? do i have to create a pull reuest? sorry if wrong place to ask


cdubz:
Correct PR is at #436.

PR #443, opened 2022-04-16
Allow end of feeding to be in the future

myxor:
This makes totally sense for us as we sometimes want to create the feeding entry in babybuddy while the feeding is still ongoing but we already know that we are going to end the feeding some short time in the future (like a few minutes).

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -361,7 +361,6 @@ class Feeding(models.Model):
 
     def clean(self):
         validate_time(self.start, "start")
-        validate_time(self.end, "end")
         validate_duration(self)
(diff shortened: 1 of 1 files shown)


cdubz:
My reasoning for this has always been to prevent _accidentally_ setting a value in the future (e.g., when entering data wearily late in the night). That said its easy enough to notice and fix an accident like that. I am gonna put this to the folks in [Gitter](https://gitter.im/babybuddy/Lobby) for opinions as well.

cdubz:
Ok this has been asked for before and there is some additional support in chat. Let’s try it 👍

PR #445, opened 2022-04-18
Fix KeyError at /children/x/reports/sleep/pattern/

myxor:
Fixes

```
  File "/app/babybuddy/reports/views.py", line 141, in get_context_data
    context["html"], context["js"] = graphs.sleep_pattern(instances)
  File "/app/babybuddy/reports/graphs/sleep_pattern.py", line 97, in sleep_pattern
    _add_adjustment(adjustment, days)
  File "/app/babybuddy/reports/graphs/sleep_pattern.py", line 194, in _add_adjustment
    days[column].append({"time": 0, "label": 0})

Exception Type: KeyError at /children/x/reports/sleep/pattern/
Exception Value: '2022-04-18'
```

which occurred with sleeping entry having start=`2022-04-17 20:58` and end=`2022-04-18 00:10`.

### DIFF
--- a/reports/graphs/sleep_pattern.py
+++ b/reports/graphs/sleep_pattern.py
@@ -190,6 +190,8 @@ def _add_adjustment(adjustment, days):
     :param blocks: List of days
     """
     column = adjustment.pop("column")
+    if not column in days:
+        days[column] = []
     # Fake (0) entry to keep the color switching logic working.
     days[column].append({"time": 0, "label": 0})
 


cdubz:
@myxor How can I reproduce the error to test this?

myxor:
@cdubz i had a sleeping entry having start=2022-04-17 20:58 and end=2022-04-18 00:10 being in timezone Europe/Berlin.
I think the crossnight entry produced the bug.
I can no longer reproduce it as i have patched my Installation already.

cdubz:
Got it! Can reproduce and this fixes it. Thanks!

PR #447, opened 2022-05-02
Update django.po

Linutux:
Formula is not a word in german. Changed to ready to use meal.


Linutux:
I updated it using the Po editor

PR #450, opened 2022-05-10
Add carousel of last 7 days of feedings to the todays feeding card

DanBeard:
My wife wanted to be able to quickly see the amount and number of feedings for the past week from the main dashboard. This review changes the feeding_day card to return a list of the last 7 days and turns the card itself into a carousel like the feeding_types card. It still only queries the DB once and then organizes/sums it in a single pass over the instances, so it should add negligible overhead. It also uses timezone aware datetimes instead of dates (to prevent incorrect sums around midnight if the local timezone is far from UTC).

 Test_card_feeding_day() and the fixture were updated. It passes but other unrelated tests are failing locally both on master and this branch. (Is this a known issue? or something up with my local environment?) Let me know what you think.


cdubz:
@DanBeard thanks for contributing! Could you please run `gulp format` on your changes so CI can get past linting and try tests?

DanBeard:
Formatted and pushed
Also here's a recording of the UI :
feedings

cdubz (changes requested):
Ok so the test fails are because of changes to the fixtures that other tests rely on. The three tests marked as failing (https://github.com/babybuddy/babybuddy/runs/6379000407) will need to updated to account for the fixture changes.

DanBeard:
Ah nice, thanks for running that. I get way more failures locally. (In things like Diaper and child tests too). I'll destroy and recreate my dev environment and see if I can get only those same failures.

cdubz:
Interesting. If you continue to get local failures post some examples and I can take a look.

DanBeard:
yeah on the most recent master I get the following test failures (see attached file).  It looks like most are the same kind of test, editing the forms. It looks like for some reason Model.refresh_from_db() isn't working as expected. Like it's still getting the old value. That last test with the date formatting in templatetags, no idea what's going on there. 

 I followed the env setup here https://docs.baby-buddy.net/contributing/development-environment/#installation with the following specs :
Ubuntu 20.04.4 LTS
npm v14.15.4 (via nvm)
Python 3.8.10
[babybuddy_error_out.txt](https://github.com/babybuddy/babybuddy/files/8672090/babybuddy_error_out.txt)

if it's ok with you, since I know these are failing, I'll skip them locally and keep the scope here to just this enhancement, and create a separate issue that you can prioritize appropriately (since maybe I'm the only one seeing it and there's just something really  weird going on with my environment? reading those tests, they seem like they should pass fine)

DanBeard:
Ok, I think I got it where it should pass. the only failures locally are those that I also see locally on master

cdubz (changes requested):
@DanBeard tests passing now!

I do have one additional CR... the "Last Feeding Method" card uses "time ago" style and I think this one should too (at least for the first two days). The first slide should be "today", the next slide "yesterday", and then additional slides can use a date but lets also add the _day_ to the date (and maybe even drop the year). I think this will make it easier for users to have a sense of what point in time they are looking at on any of the slides independently.

cdubz (approved):

cdubz:
Looks great -- thank you!

PR #463, opened 2022-05-30
Add 'max tag count' setting for quick searches

bugbountyguy:
Just a quick little thing to help with the quick tags. I found them super useful, but while adding tags, we noticed that 5 was always 1 or 2 too-short for the 'quick' suggestions. I thought this might be a nice little setting to have. It was becoming a little frustrating trying to remember the exact casing/spacing/etc we had for some of the tags while updating things :) 

However: notice the TODO I added (intentionally)... This only handles grabbing the setting from the _first user_. I could not find a clean approach or way to pass the request/session to the TagsEditor class so that I could just pull out the setting from the current user. Any suggestions on how to do this would be appreciated!

Future suggestion: Let an auto-dropdown print `most_tags` or do queries as the user types in tags.


cdubz:
@bugbountyguy are you still working on this PR?

cdubz:
> terribly sorry for the over delayed response! 

No worries 👶

I don't have strong feelings about this so I'm going to close the PR for now. But do feel free to re-open if you want to work on it again (ditto for anyone else who digs this PR up).

PR #466, opened 2022-06-02
Update Spanish translations

Alberdi:
~As always, there's a full rewrite of django.mo after exporting from POEditor :(~

Nevermind, figured that I can run `makemessages` afterwards to reorder them.


cdubz:
> Nevermind, figured that I can run makemessages afterwards to reorder them.

Nice! I didn't know that.

PR #476, opened 2022-06-06
✨ adds support for other S3-like providers

jljouannic:
Apart from AWS, there many cloud providers that provides S3-compatible object storage. To use on of these providers, an `AWS_S3_ENDPOINT_URL` variable must be specified. I propose to add support for this variable in `base.py` setting file


jljouannic:
@cdubz Good point! And done :smiley:!

PR #478, opened 2022-06-11
Use local timezone for pumping graph data

appletalk:
This corrects the pumping graph to use local timezone rather than UTC.


appletalk:
Fixed a typo, oops

cdubz:
Nice catch. Thanks for contributing!

PR #484, opened 2022-06-15
Fix dayssince sometimes being off by 1 day depending on timezone

DanBeard:
dayssince pulled UTC date for now() instead of using localtime(). This would sometimes cause all the days to be one off (e.g. "today" -> "yesterday").


cdubz:
Nice catch. Thanks!

PR #485, opened 2022-06-15
Sleep history

DanBeard:
Add a list of sleep totals for the last 7 days. Useful for seeing how well the little ones are sleeping (or not :sweat_smile:  ) . If a sleep session goes over midnight then the duration that is before midnight is added to the start day, and the portion that is after midnight is added to the end day. 

Unsure of what noun to use for "sleep" in the dashboard since "nap" has an already established meaning and this counts all sleep sessions, not just naps. Went for the simple noun "sleep"

Relies on #484 . 

sleep


core/templatetags/duration.py
```diff
@@ -103,7 +103,7 @@ def dayssince(value, today=None):
     :returns: the formatted string
     """
     if today is None:
-        today = timezone.datetime.now().date()
+        today = timezone.localtime().date()
```
cdubz, line 106: This is merged in now so you can rebase.

dashboard/templatetags/cards.py
```diff
@@ -222,41 +222,70 @@ def card_sleep_last(context, child):
 
 
 @register.inclusion_tag("cards/sleep_day.html", takes_context=True)
-def card_sleep_day(context, child, date=None):
+def card_sleep_day(context, child, end_date=None):
```
cdubz, line 225: Lets rename this card and template to just `sleep` or maybe `sleep_recent` -- you can decide (:

cdubz (changes requested):
Just a couple of small change requests for now. I'll do more testing later.

dashboard/templates/cards/sleep_recent.html
```diff
@@ -0,0 +1,54 @@
+{% extends 'cards/base.html' %}
+{% load duration i18n %}
+
+{% block header %}
+    <a href="{% url "core:sleep-list" %}">
+        {% trans "Today's Sleep" %}
```
cdubz, line 6: Lets change this is to "Recent Sleep" as well.

cdubz (changes requested):
One more small CR -- otherwise looks good!

dashboard/templates/cards/sleep_recent.html
DanBeard: Ok, Does that mean we need to go through each *.po file to update the localized String?

cdubz: Yeah but I can handle that in a separate commit.

If you want to you can run `gulp makemessages -a` and then `gulp compilemessages` but sometimes the message stuff needs some fiddling (I think it _shouldn't_ in this case).

DanBeard: ran the messages command. The below had a newline in ES (line 791 in es/django.po), so I added it in EN as well since it was throwing an error.

`"Add <samp>%(origin)s</samp> to the <code>CSRF_TRUSTED_ORIGINS</code> "
"environment variable. If multiple origins are required separate with commas.\n"`

cdubz (approved):

cdubz:
All good -- thanks!!

Issue #174, opened 2020-11-01
Issue with times around daylight savings


JoeSc:
I was just up feeding my child and got an error saying.
```
2020-11-01 01:06:00 couldn’t be interpreted in time zone America/Chicago; it may be ambiguous or it may not exist
```
The above error was generated from the '+' -> 'Diaper Change' page.

I assume this has to do with daylight savings. Since even I don't know if right now is the first or second time it's 1:06 am...

cdubz:
Oaf. Sorry to hear that. I’ll see if I can reproduce and figure out what’s happening there. Damn DST...

JoeSc:
Interestingly the "Time fill in popup" has two "01s" in it
![image](https://user-images.githubusercontent.com/367827/97805715-6e97bf00-1c1d-11eb-8690-5691b9386c18.png)
So it looks like "time fill in popup" understands the proper time,  might just be how you are interpreting the data that is submitted.

cdubz:
The error occurred for you when you got to the form, correct? I.e. not when you _submitted_ the form? I’m guessing this has something to do with logic relating to prefilling the date field values when the form is loaded.

BenjaminHae:
For me it came after submitting a timer started during the second hour between one and two.

cdubz:
Thanks @BenjaminHae. I’ll check that out as well. Looks like this is reproducible even now by selecting a time between 1 and 2a —

<img width="375" alt="PNG image" src="https://user-images.githubusercontent.com/10456740/97806835-7a31b880-1c12-11eb-9ae7-476b673f1e47.png">

Luzr:
Same issue for Chicago Timezone.
![2020-11-01_09-41-27](https://user-images.githubusercontent.com/9487415/97807460-8fb0dd80-1c26-11eb-8744-674d94f4dfa6.png)

amandadebler:
I saw the same thing during Europe's time change a few weeks ago

axilleas:
👋 any workarounds on this? Seems we cannot use a time between 2am-2:59am when time changes.

btw I'm getting this when submitting the form.

cdubz:
No workarounds I can think of. I can’t believe this bug report is a year old. I hate to say “soon” but I will try to address this… soon 😬

axilleas:
Yeah, I know :smile: No worries, the current workaround is to not file any reports during this hour :)

kdrobnyh:
[...]

PR #487, opened 2022-06-16
Fix 500 in sleep pattern

kejadlen:
I was still seeing a 500 (not sure if this is the same as #174 or not), was able to reproduce it in a test, and this fixed the issue.

Also ran Hypothesis on `sleep_pattern` in [this commit](https://github.com/kejadlen/babybuddy/commit/3011b3135f2d568dd1e3cbc00c6fdcf65ef05d18), which did produce the same failure that I saw. Didn't catch any other ones with the simple strategy though.

### DIFF
--- a/reports/tests/graphs/tests_sleep_pattern.py
+++ b/reports/tests/graphs/tests_sleep_pattern.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-from datetime import datetime
-from zoneinfo import ZoneInfo
+import datetime as dt
 
 from django.test import TestCase
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


reports/tests/graphs/tests_sleep_pattern.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from datetime import datetime
+from zoneinfo import ZoneInfo
```
cdubz, line 3: We still support Python 3.7 and 3.8 and I'd like to keep it that way for as long as those versions are supported ([for another 2+ years](https://devguide.python.org/#status-of-python-branches)). Rather than use backports, I wonder if there is some other way to configure this test?

cdubz (changes requested):

reports/tests/graphs/tests_sleep_pattern.py
kejadlen: Yup, done - switched to using `timezone` from datetime.

cdubz (approved):
Looks great -- thanks for providing a test!

PR #495, opened 2022-07-08
Adds settings file and documentation to deploy on clever cloud

jljouannic:
I have chosen to deploy babybuddy on Clever Cloud (a French "Heroku-like" cloud provider). Maybe the time I spent to write the dedicated settings file and set the right environment variables could be useful for someone else. So here comes this pull-request!


cdubz:
Nice -- happy to have more deployment methods. Thanks!

PR #500, opened 2022-07-22
Improves Quick Start Timer UX in multiple children case

jljouannic:
Quick Start Timer as it is is great to quickly start a timer as it automatically add a child relationship if there is only one child instance. But when you have 2, 3 or more child instances, you have to do a **lot** of mouse clicks to link the timer to the child you want. Here is my proposal to quickly start a timer and at the same time link it to a child:

- keep the current UX if there is only one child instance,
- on mobile, make the quick start timer icon display a dropdown with child instances to click on instead of directly start the timer:

    [Capture d’écran vidéo de 22-07-2022 16:44:24.webm](https://user-images.githubusercontent.com/862310/180464416-86b04af5-b8f4-4657-af32-b26a53fa204a.webm)
- similar experience on desktop by displaying child instances in navbar dropdown:

    [Capture d’écran vidéo de 22-07-2022 16:48:28.webm](https://user-images.githubusercontent.com/862310/180465028-79c75ca1-6c3d-4698-bbd5-0413fdb07da9.webm)

To implement this, I also had to do a fix about the way double form submission was prevented.


cdubz (approved):
Thanks for another high quality PR -- this is great!

Issue #503, opened 2022-07-29
Nap vs sleep


lukyjay:
Hi

Some of my sleep entries are being recorded as 'naps' and others as 'sleep'. It seems to depend on whether it comes from the Home Assistant addon or directly from BabyBuddy interface. Neither of these interfaces offer the option of selecting nap vs sleep as far as I can see.

The nap function appears to be pointless. What is the difference between nap and sleep?

My suggestion is to remove the nap function and only track sleep. If needed on the frontend, perhaps naps could be defined based on time of day or duration of sleep.

Any clarification would be appreciated :)

Thank you

cdubz:
Nap classification is based on the [`NAP_START_MIN`](https://docs.baby-buddy.net/setup/configuration/#nap_start_min) and [`NAP_START_MAX`](https://docs.baby-buddy.net/setup/configuration/#nap_start_max) settings for the app. #325 made the classification in to a field and there was meant to be a follow-up PR to make it togglable by the user but that work never happened.

If you look in to you'll find there is information and recommendations about number and length of naps specifically (in addition to sleep overall) based on age that may be of interest to some parents ([example article](https://www.thebump.com/a/how-many-naps-do-babies-need)).

For our part back when we used the app we received a chart with information about naps and sleep from the hospital so we used Baby Buddy to compare and sometimes inform our own decisions about, e.g., when to put a child down for a nap.

Different parents will of course have differing opinions about the relevance or usefulness of this information but my philosophy for Baby Buddy has always been to try to support as many caregivers as possible (with a sometimes not so vigilant eye on potential bloat).

Hope that clarifies a bit about why the configuration is the way it is currently.

lukyjay:
I agree naps are important and see the why every parent should know about them. I raised this issue to question the need for a seperate 'nap' function in BabyBuddy, and if it is needed, to raise some concerns on how it is currently implemented.

I'd be happy to track my naps as 'sleep' and will know that they are naps via their shorter duration.

To summarise some defects for the 'nap' implementation:

1. Sleep entries appear to default as 'naps' if there is no nap data provided (e.g. the Home Assistant addon doesn't provide any data on whether nap is 'true' or 'false', so BabyBuddy defaults all sleep entries to true (shouldn't nulls always default to false?))
2. No information in the BabyBuddy documentation on the 'nap' feature
3. No way to toggle whether sleep is a nap or sleep via the BabyBuddy timer interface
[...]

PR #504, opened 2022-07-29
Add nap toggle to sleep form and allow editing naps/sleep

lukyjay:
Hi

As per some points under: https://github.com/babybuddy/babybuddy/issues/503

Please note I am new to Python and have tested these changes but need someone to review them with a fine tooth comb in case I have made a mistake. I am still learning and appreciate any feedback and correections/improvements you can provide.

These changes add a 'napping' toggle to the sleep form and allow the user to edit whether the sleep was a nap or not. It also allows you to edit previous sleeps through the django admin interface.

Thank you

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -210,7 +210,7 @@ class NoteForm(CoreModelForm, TaggableModelForm):
 class SleepForm(CoreModelForm, TaggableModelForm):
     class Meta:
         model = models.Sleep
-        fields = ["child", "start", "end", "notes", "tags"]
+        fields = ["child", "start", "end", "napping", "notes", "tags"]
         widgets = {
             "child": ChildRadioSelect,
             "start": forms.DateTimeInput(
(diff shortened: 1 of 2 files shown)


cdubz (changes requested):
Unfortunately it's a bit more complicated than this. To have an orderly change here I think we need to do the following:

- [ ] Remove the calculated `core.models.Sleep.nap` property entirely.
- [ ] Update `core.models.NapsManager` to use `napping` instead of `nap` property.
- [ ] Update `core/templates/core/sleep_list.html` to use `napping` instead of `nap` property.
- [ ] Add a new method to `core.forms.SleepForm` to evaluate `NAP_START_MIN` and `NAP_START_MAX` and pre-select (or not) the nap checkbox in the form. Ideally we would handle this in a way that also covers the Django admin ("backend") form.
- [ ] Add a new method to `api.serializers.SleepSerializer` to do the same evaluation for data coming via the API -- setting the `napping` property to `true` or `false` based on the time and configuration _if_ it's value is not provided by the request.
- [ ] Add tests

cdubz:
Closing in favor of tracking in #513.

PR #511, opened 2022-08-03
🚸 adds a child quick switch in breadcrumb

jljouannic:
I found it tedious to consult graphs for one of my twins, then to consult the same graphs for the other. So I propose a quick switch in the breadcrumb. It is available in dashboard, timeline and reports breadcrumbs and allows to quickly switch to the "same" page but for another child

[Capture d’écran vidéo de 03-08-2022 14:59:36.webm](https://user-images.githubusercontent.com/862310/182614012-1012dd83-18d5-40fe-990c-345abe6b002b.webm)

### DIFF
--- a/core/templates/core/child_detail.html
+++ b/core/templates/core/child_detail.html
@@ -1,11 +1,13 @@
 {% extends 'babybuddy/page.html' %}
-{% load duration i18n imagekit static %}
+{% load breadcrumb duration i18n imagekit static %}
 
 {% block title %}{{ object }}{% endblock %}
 
 {% block breadcrumbs %}
     <li class="breadcrumb-item"><a href="{% url 'core:child-list' %}">{% trans "Children" %}</a></li>
-    <li class="breadcrumb-item font-weight-bold">{{ object }}</li>
+    <li class="breadcrumb-item font-weight-bold">
+        {% child_quick_switch object 'core:child' %}
+    </li>
 {% endblock %}
 
 {% block content %}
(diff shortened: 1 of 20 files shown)


cdubz:
@jljouannic thanks for another great enhancement for multi-child databases!

What do you think about using a primary color arrow only for the drop down? I think it looks a little nicer but maybe the U/X of the current design is better?

This is essentially just changing the `a class` to `btn btn-xs text-primary dropdown-toggle dropdown-toggle-split` --

Screenshot 2022-08-03 at 08-39-04 Bobby Case Baby Buddy

vs.

Screenshot 2022-08-03 at 08-39-58 Bobby Case Baby Buddy

jljouannic:
> What do you think about using a primary color arrow only for the drop down?

This is what I did first, but I found it confusing as it looked like one of the navbar menu, whereas the behaviour is different (clicking on the child name is not meant to open the dropdown). Moreover, I found it difficult to click on such a small caret on a mobile phone.

That being said, I am not happy with my design either.

Maybe with an icon like [this one](https://icons.getbootstrap.com/icons/caret-down-square/)? What do you think?
image

cdubz:
Ok -- good point about that being confusing. It's easy to think about when focused on it but yeah it makes it look like the whole name will drop down.

I think that icon would be a good compromise.

jljouannic:
It's done!

cdubz:
Looks great! Thanks, again, for all these U/X improvements for multi-child databases.

PR #519, opened 2022-08-09
Install GNU gettext at gitpod startup

amorphobia:
It's handy to use GitPod to update translations. So I would like to have `gettext` pre-installed for GitPod sessions. Per my experience, installing `gettext` in GitPod would not take much time.


cdubz:
Nice. Thanks for catching this!

PR #524, opened 2022-08-25
Extra API endpoints and login qr-code for app integration

MrApplejuice:
Hello there, long time no see! I am still booking progress on the [Android-app](https://github.com/MrApplejuice/BabyBuddyAndroid/) I am developing for BabyBuddy and was thinking about ways to make the login more convenient.

I understand that this PR is coming out of nowhere and was born out of the needs of some other project. I thought it was simple, but it got complex. I think the changes/additions might be handy for other projects as well. Please read this a proposal, even though there is no issue for this. The PR is not finished yet and lacks, mainly tests, I think.

# Background

My app currently uses the API-key to authenticate with a running baby buddy webapp, but I was facing some "convenience" and stability issues:
- Logging in via the app is kind of cumbersome, requiring three separate entries to be made: See https://github.com/MrApplejuice/BabyBuddyAndroid/blob/master/doc/images/login_screen.png . Especially the URL-part is a bit - weird - for non-technical users, I found. While technically, I could reduce the number of login entries by using the API-Key directly, that seems even more alien for users to me. What the app does now is login with username & password and then parses the API-Key from the user's settings HTML, which brings me to the next issue:
- There is no API-endpoint for obtaining the API-key reliably if a session has been initiated through the login mechanism. 

# What was added/changed?

I added a api-backend `api/profile` that allows obtaining some information from the "settings" page and about the logged in user (especially the `api_key` for me to grab in the app):

```
GET /api/profile

HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "user": 1,
    "username": "admin",
    "first_name": "First name",
    "last_name": "",
    "email": "[no@localhost.local](mailto:no@localhost.local)",
    "staff": true,
    "language": "en-US",
    "timezone": "UTC",
    "api_key": "fb57a258f9933b8b7ca0127477dc9a20cce8de89"
}
```

And then to make it more convenient in the future to login to baby buddy on mobile apps, I added a "Login QR Code" that contains everything required to login to the website:
- The website address
- The API key

In a code like this one: 

```
BABYBUDDY-LOGIN:{"url":"http://192.168.0.10:8000/","api_key":"80c6e4b8b94b09a5167a5a0fc09c3ef6455a113c"}
```

image

# Some collateral changes/fixes

Added library [qrcode](https://pypi.org/project/qrcode/) to the dependencies.

The api-root view now also enumerates non-ModelView pages (schema and profile):

```
{
    "bmi": "http://192.168.0.10:8000/api/bmi/",
    "changes": "http://192.168.0.10:8000/api/changes/",
    "children": "http://192.168.0.10:8000/api/children/",
    "feedings": "http://192.168.0.10:8000/api/feedings/",
    "head-circumference": "http://192.168.0.10:8000/api/head-circumference/",
    "height": "http://192.168.0.10:8000/api/height/",
    "notes": "http://192.168.0.10:8000/api/notes/",
    "pumping": "http://192.168.0.10:8000/api/pumping/",
    "sleep": "http://192.168.0.10:8000/api/sleep/",
    "tags": "http://192.168.0.10:8000/api/tags/",
    "temperature": "http://192.168.0.10:8000/api/temperature/",
    "timers": "http://192.168.0.10:8000/api/timers/",
    "tummy-times": "http://192.168.0.10:8000/api/tummy-times/",
    "weight": "http://192.168.0.10:8000/api/weight/",
    "profile": "http://192.168.0.10:8000/api/profile",
    "schema": "http://192.168.0.10:8000/api/schema"
}
```

# TODOs

- [x] Add tests


cdubz:
Oh wow! This is very cool. I might be a little slow on the review here but this definitely seems worth consideration. And I'm very excited that you have kept at the Android app!

MrApplejuice:
Nice to hear that the changeset is to your liking!

There is one strategic consideration wrt to the qrcode generation code. Right now, it is quite a custom solution with a lot of template-tag code. On the other hand there is this project: https://django-qr-code.readthedocs.io/ which seems well maintained and should also do the trick. I think that the current "custom" solution can be replaced by django-qr-code which will reduce the size of the changeset considerably.

cdubz:
@MrApplejuice yeah from a quick  scan I was wondering about all that QR code stuff. Do you want to try out django-qr-code before I review?

MrApplejuice:
Yeah, let me have another look at the qr-code stuff. I opened the PR as a draft to check if the idea is sane.

I ran into some issues with existing QR-code libraries because I wanted to leverage django templating for the contents of the QR code itself, but I find this solution not super readable either... I will try another solution first, I think,

api/views.py
```diff
@@ -113,3 +116,16 @@ class WeightViewSet(viewsets.ModelViewSet):
...
+
+
+class ProfileView(views.APIView):
+    basename = "profile"
+    queryset = babybuddy_models.Settings.objects.all()
+    serializer_class = serializers.ProfileSerializer
+
+    def get(self, request):
+        settings = get_object_or_404(
+            babybuddy_models.Settings.objects, user=request.user
```
MrApplejuice, line 128: Custom get required to query `request.user`

MrApplejuice:
Alright, I reworked this to use django-qr-code instead. Little caveat - it looks more ugly now:

grafik

However, I learned that this (unlike my own implementation) actually is more standard-compliant. The qr-code standard calls for clearance around the QR-code itself if 4 "units" or so (the boxes the QR-code consists of), and this seems to be more compliant with its fat border.

I added tests as well now, and also added the `username` to the `api/profile` endpoint.

babybuddy/templates/babybuddy/login_qr_code.txt
```diff
@@ -0,0 +1,4 @@
+{% load i18n widget_tweaks babybuddy_tags qr_code %}
+{% url 'babybuddy:root-router' as relative_root_url %}
+{% make_absolute_url relative_root_url as absolute_root_url %}
+BABYBUDDY-LOGIN:{"url":"{{ absolute_root_url }}","api_key":"{{ user.settings.api_key }}"}
```
cdubz, line 4: I wonder... is there any way we can make this _also_ work with a regular browser? Or is that not a thing? I'm unsure how QR codes work haha. It seems like it would be convenient as a way to easily log in via a mobile web browser from a logged in desktop, for example.

Based on this random page it seems like this should be possible but I can't find any documentation about how to have a URL "fallback": https://qr-creator.com/detect-iOS-Android.php

If that's not possible we'll need to think on a better way to explain what the code is for... maybe just calling the field "Android App login code"? Seems kind of verbose...

cdubz (changes requested):
Looks much better with the Django app! A few items in review comments.

babybuddy/templates/babybuddy/login_qr_code.txt
MrApplejuice: Hey... took some time for me to write a complete answer on this topic specifically. It namely hits home for me wrt a problem I am facing with the Android app as well:

Right now the `api_key` actually only seems to give access to the `/api`-views. I think this is not too crazy, actually. The `api_key` is for using that API. What I am trying to do with the Android app is to make using timers very convenient and send everyone to the actual website for everything else. For this I need to forward people to the website, that I cannot show without a valid user session. User sessions cannot be generated with the `api_key`.

So the bottom-line seems to be: "Yes, this does not exist" for the regular website - unfortunately. At least not with the current authentication mechanisms.

--------------------

I really would like to do something like that in the future using the `api_key` for my app. So I was thinking myself about this. "Funnily" enough, I am maintaining a plugin for wordpress that does exactly that: https://github.com/MrApplejuice/wp-autologin-links (kind of dormat right now) - it adds a token-login (like the api-key is, in principle).

From my experience with the plugin, I would however suggest that not to do this. It is super-unsafe, generally. Typically one has to embed the login code in the url as a get-parameter, because there is no other safe way of embedding the code in a plain-text representation (like a QR-code is), but then again it is super easy to share such a link by accident for a user and pretty much leak his/her password and can get full access to his/her account. This is different from the API, which allows a lot of access, but not, for example, to change one's password. Such a code is convenient, but only if the website is designed for that. I do not know that this would be a "good feature" in that sense.

What I would suggest instead is to create a special backend on the API-side that will allow generating actual user sessions using a POST message, for example. This will open up an avenue for developers to open user sessions without storing a user password for authentication purposes. Since we are talking about POST-requests, the secret `api_url` can remain a header and cannot be encoded in the URL anyway (security feature), and it would be possible to configure that special endpoint in such a way that allows admins to lock down that potential vulnerability if is undesirable.

I think this is my stance on this. Token authentication is handy but not super-safe. There are better ways that integrators can use and users should probably stick to password authentication and store their passwords in a password manager which is just as convenient as a login-token.

api/serializers.py
```diff
@@ -268,3 +269,45 @@ class WeightSerializer(CoreModelSerializer, TaggableSerializer):
...
+        return self.instance.user.last_name
+
+    def get_email(self, value):
+        return self.instance.user.email
+
+    def is_staff(self, value):
+        return self.instance.user.is_staff
+
+    class Meta:
+        model = babybuddy_models.Settings
```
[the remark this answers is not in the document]
MrApplejuice: > UserSerializer

Yeah, no problem. I will remove that one :+1:

> DRF nested relationship

I will need to do some research on how that one works! Sounds neat and handy though - if it works!

api/views.py
MrApplejuice: Will check!

api/serializers.py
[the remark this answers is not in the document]
MrApplejuice: Works, changed the layout of the `/api/profile` endpoint a bit, but does not matter I think:

~~~~~~~~~~~
{
    "user": {
        "id": 1,
        "username": "admin",
        "first_name": "First name",
        "last_name": "",
        "email": "[no@localhost.local](mailto:no@localhost.local)",
        "is_staff": true
    },
    "language": "en-US",
    "timezone": "UTC",
    "api_key": "1234567890..."
}
~~~~~~~~~~~

MrApplejuice:
I left the Profile-implementation to enumerate Settings rather than Users which might be a bit weird, but was a little bit easier, because Settings link to the Token AND the User. Enumerating both from the Settings was a little bit easier. Probably does not matter too much, but it leads to this layout now:

~~~~~~~~~~~
{
    "user": {
        "id": 1,
        "username": "admin",
        "first_name": "First name",
        "last_name": "",
        "email": "[no@localhost.local](mailto:no@localhost.local)",
        "is_staff": true
    },
    "language": "en-US",
    "timezone": "UTC",
    "api_key": "6ce48e3f5f20a767c5c7356f89637484cfd94585"
}
~~~~~~~~~~~

api/serializers.py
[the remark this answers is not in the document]
cdubz: Based on the data model that's exactly what I would expect so 👍

babybuddy/templates/babybuddy/login_qr_code.txt
cdubz: > It is super-unsafe, generally. 

Yeah. I tend to think that's OK for an app like this but I suppose that's definitely not true of all users (or even a majority, who knows!). So maybe this would just be a nice optional feature for some (using e.g. https://github.com/aaugustin/django-sesame) -- separately.

> What I would suggest instead is to create a special backend on the API-side that will allow generating actual user sessions using a POST message, for example. This will open up an avenue for developers to open user sessions without storing a user password for authentication purposes.

Whats the use case for this? I'm unclear on that.

cdubz: Regardless I still want to make clear to the user what this QR code is for. As it currently it the "Login QR code" title is kind of confusing. Maybe we should generate the code on a separate page and link to it from somewhere instead? Or put this in a separate "Android" fieldset?

cdubz: @MrApplejuice any thoughts on follow up here? I can fiddling with this if you've got more pressing issues right now 👶

MrApplejuice: Hey... whoops, sorry. Somehow I marked this as "done" in my head. 

> Regardless I still want to make clear to the user what this QR code is for. As it currently it the "Login QR code" title is kind of confusing.

"Hiding" the page somewhere else could indeed be a nice option. I quickly checked how Telegram does it which I had ready on my phone. It has a menu-item "Link a new device" under the menu option "Devices". Not really something that babybuddy can copy 1:1 because it would mean that individual devices get their own API keys (though that would be cool, I guess).

Maybe something like this (quick mock-up):

image

And then on the new page you get presented with the API-URL and the QR-code maybe?

cdubz: > Hey... whoops, sorry. Somehow I marked this as "done" in my head.

Haha I did too.

> Maybe something like this (quick mock-up):

That looks fine -- lets do it!

MrApplejuice: Quick expectation management: I will need another week to implement this. Hope that is no problem?!

cdubz: Absolutely. Baby Buddy has no deadlines 😀

MrApplejuice:
Okay. I reworked the pages. Here a few screenshots:

grafik

grafik

Two open small questions:
- The sentence I wrote "This page allows you to gain access to babybuddy from a third party app or device." feels clunky. Do you as a native speaker (I assume?) have a better suggestion? I tried different phrases but all get confusing with personal pronouns pointing to all the wrong things... Maybe some out-of-the-box thinking is required... 
- I left the API-key on the settings-page, but in theory that is not needed anymore. It can now just as well be placed on the "Add a device" page, which would remove the entire "API" section from the Settings-screen. What do you think?

cdubz:
> The sentence I wrote "This page allows you to gain access to babybuddy from a third party app or device." feels clunky. Do you as a native speaker (I assume?) have a better suggestion? I tried different phrases but all get confusing with personal pronouns pointing to all the wrong things... Maybe some out-of-the-box thinking is required...

How about, "Use Baby Buddy from other apps or devices."

> I left the API-key on the settings-page, but in theory that is not needed anymore. It can now just as well be placed on the "Add a device" page, which would remove the entire "API" section from the Settings-screen. What do you think?

I'm not sure I like the idea of moving the API key as well -- is it required for the Android app log in, too? I wonder if there is a more generic name we could use for this section if we do have both here...

- External Access,
- External Authentication,
- Devices & APIs,
- Authentication Methods,
- something else?

MrApplejuice:
> I'm not sure I like the idea of moving the API key as well -- is it required for the Android app log in, too? 

No not at all. Was just thinking of de-duplicating the interface. Having the API-key on the "Add a device" page makes sense, because it _could_ be another option for authenticating an external app. But if the API-keyt is there on the "Add a device" page, maybe it was not needed on the Settings-page either. Not a requirement, just a suggestion.

> How about, "Use Baby Buddy from other apps or devices."
and
>  [...] more generic name

I reworked the interface a bit and actually removed the sentence I added before again. I think using a more generic name as the sub-section title is quite appropriate. Here:

grafik

The interface looks a little bit barren, but context for this page can be given to a user by an app that wants to gain access. Also in the future, it could maybe be expanded with with more options if desired (like multiple API-keys for different apps, for example). I think having a page for this stuff makes sense even if it is empty right now.

MrApplejuice:
> Fix breadcrumbs-commit

Just realized in the screenshot that I did not change the breadcrumbs - that one if fixed now as well! :D

babybuddy/views.py
```diff
@@ -147,7 +149,16 @@ def post(self, request):
         return render(request, self.template_name, {"form": form})
 
 
-class UserSettings(LoginRequiredMixin, View):
+class RegenerateApiKey:
+    def handle_api_key_post(self, request) -> Boolean:
```
cdubz, line 153: Should this be just `bool`? I noticed `Boolean` here is imported from `xmlrpc.client` and that seems odd...

cdubz (changes requested):
Just one last Q re: a typing...

babybuddy/views.py
MrApplejuice: Whoops. You are right. Carry-over from my Android Java/Kotlin project, where the types have different names. Auto-import was _too powerful_ and found a type for "Boolean" anyway :laughing:

MrApplejuice: Also noticed that this does not need to be a class-method anymore. Refactored it to be a plain method.

cdubz (approved):

cdubz (approved):
We did it!

MrApplejuice:
Yeay, thanks a lot! :-)

Issue #517, opened 2022-08-05
Allow REMOTE_USER authentication 


EnsuingRequiem:
I'd like to see the RemoteUserMiddleware and RemoteUserBackend be able to be added via environment variable similar to how Tandoor Recipes implemented it (https://github.com/TandoorRecipes/recipes/commit/55777fd948e1fb18683ac833ba72cdc78c95ec75) in order to allow for forward-authentication/sso when setup behind a reverse proxy.

cdubz:
Should be doable! Happy to take PRs or I’ll take a look at this eventually.

EnsuingRequiem:
Cool! Since I have a bit of a guide to work from, I'll see if I can put together a PR. I should probably make sure that I have it working properly on my Tandoor container to have a comparison.

PR #531, opened 2022-09-17
Add forward auth by way of remote user

EnsuingRequiem:
Implements feature requested in #517 

I tested the final state (containerized docker deployment) by modifying the LinuxServers.io Dockerfile to use my fork with the committed changes. The implementation works wonderfully. The username being passed in the header must exist in the system beforehand. If the header isn't received, authentication falls back to the login page.

### DIFF
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -10,10 +10,6 @@ from os import getenv
 from django.contrib.auth.middleware import RemoteUserMiddleware
 
 
-class CustomRemoteUser(RemoteUserMiddleware):
-    header = getenv("PROXY_HEADER", "HTTP_REMOTE_USER")
-
-
 def update_en_us_date_formats():
     """
     Update the datetime formats for the en-US locale. This is handled here and
@@ -149,3 +145,7 @@ class RollingSessionMiddleware:
             else:
                 request.session["session_refresh"] = int(time.time())
         return self.get_response(request)
+
+
+class CustomRemoteUser(RemoteUserMiddleware):
+    header = getenv("PROXY_HEADER", "HTTP_REMOTE_USER")
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -1,4 +1,5 @@
-import time
+from os import getenv
+from time import time
 
 import pytz
 
@@ -6,7 +7,6 @@ from django.conf import settings
 from django.utils import timezone, translation
 from django.conf.locale.en import formats as formats_en_us
 from django.conf.locale.en_GB import formats as formats_en_gb
-from os import getenv
 from django.contrib.auth.middleware import RemoteUserMiddleware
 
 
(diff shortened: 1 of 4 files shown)


babybuddy/middleware.py
```diff
@@ -6,6 +6,12 @@
 from django.utils import timezone, translation
 from django.conf.locale.en import formats as formats_en_us
 from django.conf.locale.en_GB import formats as formats_en_gb
+from os import getenv
+from django.contrib.auth.middleware import RemoteUserMiddleware
+
+
+class CustomRemoteUser(RemoteUserMiddleware):
```
cdubz, line 13: Could you move this class down to the bottom of the file? Just a personal preference thing to have classes after functions.

babybuddy/settings/base.py
```diff
@@ -12,6 +12,8 @@
 
 load_dotenv(find_dotenv())
 
+REVERSE_PROXY_AUTH = bool(int(os.getenv("REVERSE_PROXY_AUTH", False)))
```
cdubz, line 15: Is there a specific reason to use `bool` and `int` here? For example below for `DEBUG` we use `bool(strtobool(os.environ.get("DEBUG") or "False"))`. Can't remember why and don't know if one is better than the other. Just curious...

babybuddy/settings/base.py
```diff
@@ -143,6 +146,12 @@
 
 LOGOUT_REDIRECT_URL = "babybuddy:login"
 
+# Only include when REVERSE_PROXY_AUTH env is 1
```
cdubz, line 149: I'm a bit confused here -- up at L67 this middleware is added to the end of the of `MIDDLEWARE` but not down here it's being added again? I assume we don't want the L67 change, right?

cdubz (changes requested):
Thanks for working on this! A few CRs in this review. Also, how would you recommend testing this? Both manually and ideally we include an automated test with the PR.

babybuddy/settings/base.py
EnsuingRequiem: Since I was largely referencing how Tandoor implemented it, there's not a preference. This is where my ignorance shows; if environment variables are always stored as strings, then sticking with strtobool would be better for consistency and wouldn't expect it to impact how a docker compose file is formatted.

I'll give it a try locally and see how it works.

babybuddy/middleware.py
EnsuingRequiem: Will do.

babybuddy/settings/base.py
EnsuingRequiem: You're correct, I'll be removing the L67 change.

EnsuingRequiem:
> Thanks for working on this! A few CRs in this review. Also, how would you recommend testing this? Both manually and ideally we include an automated test with the PR.

For testing, I'm not sure since I feel like it leans towards an integration test of sorts. Technically, when run locally, a header can be spoofed (this the warning in the docs). I'll take a look at the tests and see if I can grok it; I'm not experienced in writing tests for python.

cdubz:
> For testing, I'm not sure since I feel like it leans towards an integration test of sorts. Technically, when run locally, a header can be spoofed (this the warning in the docs). I'll take a look at the tests and see if I can grok it; I'm not experienced in writing tests for python.

No worries if you don’t want to work out the automated tests. If you can tell me how to setup a manual test I can probably work it out from there. Or at the very least do my own manual test to verify the functionality.

EnsuingRequiem:
> > For testing, I'm not sure since I feel like it leans towards an integration test of sorts. Technically, when run locally, a header can be spoofed (this the warning in the docs). I'll take a look at the tests and see if I can grok it; I'm not experienced in writing tests for python.
> 
> No worries if you don’t want to work out the automated tests. If you can tell me how to setup a manual test I can probably work it out from there. Or at the very least do my own manual test to verify the functionality. 

Normally if you were to `curl` or `wget` against the application with no cookies, you'd get directed to the login page. If, instead, you `curl --header "Remote-User: validuser"` or `wget --header="Remote-User: validuser"`, then you should get the requested page without being redirected. The `Remote-User` is not case sensitive and the `-` could be `_` due to (apparently) how python normalizes the headers. The `HTTP_` part is stripped and indicates it originated in the HTTP request. If the `PROXY_HEADER` setting doesn't start with `HTTP_` then the `curl` or `wget` requests with headers won't work.

cdubz:
Looks good. Thanks for contributing!

PR #536, opened 2022-10-04
470 - "Recently used" shouldn't appear if no tags were recently used

### LINKED ISSUE #470: "Recently used" shouldn't appear if no tags were recently used
It looks a bit strange if there are no tags:
Screenshot 2022-06-03 at 20 06 13

Also, the text below could change to reflect that state.

earthcomfy:
Closes #470 

I have tried looking for a couple of solutions to avoid using inline CSS. However, since the logic is very generic, using inline CSS is the only way that worked. If the [div](https://github.com/babybuddy/babybuddy/blob/master/babybuddy/templates/babybuddy/form_field.html#L44) that showed `help_text` was in the same file as the [widget tag editor](https://github.com/babybuddy/babybuddy/blob/master/core/templates/core/widget_tag_editor.html#L27), I could have conditionally added a class to the div and hide it inside the custom CSS files, but that's not the case. Anyways maybe I'm missing something here. I would like to know if there is another better way.

### DIFF
--- a/core/templates/core/widget_tag_editor.html
+++ b/core/templates/core/widget_tag_editor.html
@@ -24,13 +24,21 @@
                 <button class="btn btn-outline-primary bg-dark btn-add-new-tag" type="button">{% trans "Add" %}</button>
             </div>
         </div>
-        <span>{% trans "Recently used:" %}</span>
-        {% for t in widget.tag_suggestions.quick %}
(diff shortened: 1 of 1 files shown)


cdubz:
Ok -- I see the issue here. We can live with the bit of inline CSS for now 😄 Thanks!

earthcomfy:
> Ok -- I see the issue here. We can live with the bit of inline CSS for now smile Thanks!

Great! Thank you

PR #538, opened 2022-10-11
#468  Minutes instead of minutos in Spanish

### LINKED ISSUE #468: Minutes instead of minutos in Spanish
When using the app in Spanish, the "since" times in the home cards aren't properly translated and display "minute(s)" instead of "minuto(s)".

I think this broke around version 1.10 and I don't think we have anything wrong in the Spanish django.po/mo files, but I'm probably wrong.

Screenshot 2022-06-03 at 19 54 15

jmunoz94:
Closes #468 

# Background

Users noticed that in the home cards in the Spanish version instead of "minutos" babybuddy shows "minutes". Upon further investigation, we noticed that "minuto" is also misspelled as "minutos" in Spanish and Portuguese is also affected by the bug.

The root cause of this issue are incorrect translations in Django. PO files for Spanish and Portuguese use either the wrong format or have been wrongly translated. As a workaround, it was suggested that we use `duration_string`.

Since `duration_string` string expects a `timedelta`, a new filter to get the duration as a `timedelta` was added (since I could not find an existing, straightforward function to use, the closes was `timesince` which returns a `str`).

# Changes

* Add `deltasince`, a new filter that calculates duration and returns a timedelta.
* Update *_last.html templates to use `deltasince` followed by `duration_string` (with `m`inutes precision) instead of `timesince`.
* Add a test for `deltasince`.
* Fix a typo in `duration.py`.

Spanish:
image

Portuguese:
image

Italian (this one already worked, adding it as a proof that other languages did not break):
image

### DIFF
--- a/core/templatetags/duration.py
+++ b/core/templatetags/duration.py
@@ -33,7 +33,7 @@ def child_age_string(birth_date):
 def duration_string(duration, precision="s"):
     """
     Format a duration (e.g. "2 hours, 3 minutes, 35 seconds").
-    :param duration: a timedetla instance.
+    :param duration: a timedelta instance.
     :param precision: the level of precision to return (h for hours, m for
                       minutes, s for seconds)
     :returns: a string representation of the duration.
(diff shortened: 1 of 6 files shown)


cdubz (approved):
Thanks, again, for your deep research and work on this! Really appreciate getting this fixed.

PR #546, opened 2022-10-17
resolves #523

### LINKED ISSUE #523: Pumping graph entries do not line up with totals (time zone issue?)
Hello,

On the Pumping graph specifically, the 8th entry is invisible even though the daily total is correct.image

garyglatfelter:
The issue was due to the last item of the _amounts_ list not being accessed in the for loop that creates plotly bar graph objects. 

If you would like I could write some test cases for this, but I'm not entirely sure how I would write test cases to check the returned plotly figure. I'm more than willing to look into it, but I've never used gulp or plotly before so I'd have to spend some more time to become familiar with them.

Fixes #523

### DIFF
--- a/reports/graphs/pumping_amounts.py
+++ b/reports/graphs/pumping_amounts.py
@@ -43,7 +43,7 @@ def pumping_amounts(objects):
         index_x += 1
 
     traces = []
-    for i in range(0, len(amounts) - 1):
+    for i in range(0, len(amounts)):
         traces.append(
             go.Bar(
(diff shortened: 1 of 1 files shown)


cdubz:
Huh... it appears I have misunderstood `range` for a long time... I wonder if there are other bugs like this scattered about 😬 

> If you would like I could write some test cases for this, but I'm not entirely sure how I would write test cases to check the returned plotly figure. 

No worries. We don't have a pattern for testing graph output.

PR #549, opened 2022-10-18
Create VS Code Devcontainer

jcgoette:
Simple VS Code devcontainer setup to simplify setting up dev environment. Also updated documentation re: this.


cdubz:
Hmmmmm I like the idea but I think I would prefer to see an IDE-agnostic version. Maybe I only feel this way because I am not a VSCode user myself 😬 Laraval has [Sail](https://github.com/laravel/sail)... I wonder if there is something similar for Django?

cdubz:
@jcgoette 👋 Still look to contribute to this branch? I'm open to finalizing and merging it if there seems to be nothing more generic around for Python+Django.

jcgoette:
Honestly, I haven't thought about this in a while. But, just did some quick web searching and found [some](https://www.untangled.dev/2020/06/06/docker-django-local-dev/) [articles ](https://medium.com/nerd-for-tech/how-to-setup-headless-dev-environment-with-django-and-docker-7495bd79b1cb)that give some guidance on IDE-agnostic Docker dev. I can pick this back up in the near future.

barakplasma:
I had an idea about re-useable devcontainers: what about extending / reusing the github actions / dokku for local development? (using the buildpack already used by the heroku / dokku / github action flow). Also wondering if maybe https://github.com/nektos/act fits in somehow

PR #550, opened 2022-10-20
feat: minify vendor.js

### LINKED ISSUE #492: Improve asset minification
### Discussed in https://github.com/babybuddy/babybuddy/discussions/491

<div type='discussions-op-text'>

<sup>Originally posted by **gbballpack** June 28, 2022</sup>
Using Cloudflare with Baby Buddy, Baby Buddy running inside of Docker with an external reverse proxy, Cloudflare is able to minify about 80 KiB from the vendor JS file.

Would it be possible to have everyone run a minified vendor JS file? Minified files generally run quicker and can reduce power consumption when on mobile compared to non-minified JS files.

I am curious if it is possible to share the benefits of the minified JS with those not using Cloudflare?

Before and after screenshots are attached, using https://web.dev/measure to verify the Cloudflare JS minification.
IMG_1425
IMG_1426</div>

mikejgray:
Closes #492

Please let me know if this isn't the right file or you'd prefer some other process. Thank you!


cdubz (changes requested):
@mikejgray thanks for opening a PR -- it looks like this is a minified version of the existing `vendor.js` file but what we really need to do to address #492 is come up with a _process_ for automatic minification.

All of the content in the `static` folder is generated by the [`gulp updatestatic`](https://github.com/babybuddy/babybuddy/blob/master/gulpfile.js#L408) command. Improving minification will most likely involve changes to the [`scripts`](https://github.com/babybuddy/babybuddy/blob/master/gulpfile.js#L181) command that is run as part of that gulp series.

It's possible that [gulp-minify](https://www.npmjs.com/package/gulp-minify) could be used here fairly easily to make a big difference. The plugin documentation is pretty clear on how it can help.

For future updates I also recommend you note in a comment the current size of the relevant files and the updated size that we'll get if the PR is merged. That will help make review easier.

mikejgray:
Will do - I'm new to gulp, so it may be a moment before I get a revision out.

cdubz:
No worries! Happy to help if you get stuck anywhere.

mikejgray:
Alright, I'm reasonably sure this was the right place to put it, but had a bit of trouble testing locally. If this isn't correct please let me know and I'll adjust as needed. If it is and you'd like to apply it to other files, I can do that too.

For reference, the original vendor.js file:

```bash
❯ du -h vendor.js
896K	vendor.js
```

cdubz (changes requested):
Looks right! Lets get the others minified as well and then I'll do some functional testing.

cdubz (changes requested):
Changes look good. To finalize you'll want to run `gulp updatestatic` and commit all the changes. If you're having trouble with gulp let me know and I can tack that one. It can be a PITA.

mikejgray:
@cdubz I tried running `npx gulp updatestatic` but it doesn't set up a virtualenv and I didn't have `black` installed in my global environment. Went ahead and installed it because it's never bad to have on hand.

```bash
❯ du -h ./babybuddy/static/babybuddy/js/vendor.js
1.2M    ./babybuddy/static/babybuddy/js/vendor.js
```

Upon further investigation it looks like the code is minifying, dropping it into the file, and then dropping the non-minified contents. I'll make some changes and get it to only drop the minified contents.

cdubz:
Interesting. That syntax looked correct to me but to fair I have not tested. Let me know if you have any trouble sorting it and I can take a look at well.

mikejgray:
@cdubz I tried a few different approaches and nothing seems to be working. If you add `minify()` to the `scripts()` function it seems to remove the script it's attached to altogether. Not sure what's going on here.

cdubz:
OK so it looks like running `minify` _before_ `concat` causes it to only minify the first file in the group. Running it _after_ `contact` looks like it works correctly --

```shell
-rw-r--r--   1 wellc  staff   2.8K Oct 22 09:04 app-min.js
-rw-r--r--   1 wellc  staff   7.3K Aug  4 08:15 app.js
-rw-r--r--   1 wellc  staff   1.2M Oct 22 09:04 graph-min.js
-rw-r--r--   1 wellc  staff   3.4M Oct 22 08:19 graph.js
-rw-rw-rw-   1 wellc  staff   4.7K Oct 22 09:04 tags_editor-min.js
-rw-r--r--   1 wellc  staff    12K Aug  3 19:51 tags_editor.js
-rw-r--r--   1 wellc  staff   358K Oct 22 09:04 vendor-min.js
-rw-r--r--   1 wellc  staff   899K Oct 22 08:19 vendor.js
```

However, this being gulp... there is a still issue in that `minify` takes a much longer time and apparently the existing gulp configuration is not properly waiting for it to finish. Looking in to that as well now...

cdubz:
Ok! Sorry to have turned this is PR in to a major refactor of the gulp config 😬

There were lots of broken things in this setup... real fun finding documentation and information for packages that were all abandoned 4+ years ago, hah.

Anyway scripts should be minimized now 🎉

cdubz (approved):

mikejgray:
Gulp seems like a real hoot...glad to know it wasn't the gulp-minify package itself. I was going nuts trying to figure out why it was removing the files when everything looked like it should work!

Sorry to add to your plate, but hopefully this boosts performance for folks. Wish I knew about this product back when my kids were babies!

cdubz:
Yeah. Some day the much larger task is getting around to switching to webpack or rollup or parcel or whatever. But gulp has been solid all these years so 🤷

PR #551, opened 2022-10-20
544 - Improve user permissions structure

earthcomfy:
Let me know if this approach is what we want so that I can also update the user form (GUI). Thanks.


cdubz (changes requested):
@earthcomfy this looks like the right direction! A couple of thoughts:

1. Lets make the `staff` user _also_ `superuser`. That will essentially represent the highest level of permissions/access. For that I don't think we'd even need a group -- the user would just have `staff` and `superuser` status.
2. Lets default to `standard`. I figure it's the more common use case and one less point of potential confusion for new users.
3. I'm assuming we'll create the groups and assign appropriate permissions in migrations, just FTR.

Do you want to tackle the frontend work on this PR as well or just CLI here and frontend in a second PR?

earthcomfy:
> @earthcomfy this looks like the right direction! A couple of thoughts:
> 
> 1. Lets make the `staff` user _also_ `superuser`. That will essentially represent the highest level of permissions/access. For that I don't think we'd even need a group -- the user would just have `staff` and `superuser` status.
> 2. Lets default to `standard`. I figure it's the more common use case and one less point of potential confusion for new users.
> 3. I'm assuming we'll create the groups and assign appropriate permissions in migrations, just FTR.
> 
> Do you want to tackle the frontend work on this PR as well or just CLI here and frontend in a second PR?

1. I see, I will add `is_superuser` arg and remove `staff` from the group options
2. Great
3. Yeah, that is what I thought too.

Let me tackle the frontend in another PR.

cdubz (approved):

cdubz (changes requested):
This looks good! Even if we separate the FE in to another PR we'll need to have the migrations on this so the groups exist for the command.

babybuddy/management/commands/createuser.py
```diff
@@ -51,6 +51,13 @@ def add_arguments(self, parser):
             dest="password",
             help="Specifies the password for the user. Optional.",
         )
+        parser.add_argument(
```
cdubz, line 54: I prefer we abstract away the fact that these are groups at all. From the perspective of a Baby Buddy user I just want to offer the three "types" of users. So instead of a `--groups` flag that could theoretically take _any_ group name lets just have two (exclusive) flags:

- `--read-only`
- `--staff`

If the `--read-only` flag is set the user is added to the `read_only` group and if the `--staff` flag is set the user is made staff and superuser (only one flag should be settable at a time). If neither flag is set the user gets added to the `standard` group.

cdubz (changes requested):
Sorry for the review delay! This all looks good except I have one final request about how we set up the command line arguments.

earthcomfy:
> Sorry for the review delay! This all looks good except I have one final request about how we set up the command line arguments.

Hi, it's alright. I have updated the arguments as per the request. Let me know :)

cdubz (approved):

cdubz:
Merging!

FYI I added two issues related to this one -- #562 is an apparent bug setting the user password using the argument (the big exists in `master` as well) and #563 is just the follow up issue for adding these features to the users management form.

Thanks for your work on this!

earthcomfy:
> Merging!
> 
> FYI I added two issues related to this one -- #562 is an apparent bug setting the user password using the argument (the big exists in `master` as well) and #563 is just the follow up issue for adding these features to the users management form.
> 
> Thanks for your work on this!

Great, you are welcome.

PR #561, opened 2022-10-31
Add condition on tummytime-duration.py in order to display hours when…

Anoerak:
… necessary


cdubz:
Looks good. Thank you!

PR #564, opened 2022-10-31
562 - Fix empty password bug in createuser command

### LINKED ISSUE #562: Passwords set by non-interactive createuser don't work
I discovered this during testing for #544 --

When using the `createuser` interactively I can set the password for the user and it works.

```shell
~ > pipenv run python manage.py createuser --username staff --is-staff
Loading .env environment variables...
Password: 
Password (again): 
User staff created successfully.
```

However, when creating a user with a password set using the command's `--password` flag I am not able to log in as the user as the password is apparently incorrect.

```shell
~ > pipenv run python manage.py createuser --username staff2 --password 2SAzqAQq --is-staff 
Loading .env environment variables...
User staff2 created successfully.
```

earthcomfy:
Closes #562

### DIFF
--- a/babybuddy/management/commands/createuser.py
+++ b/babybuddy/management/commands/createuser.py
@@ -71,7 +71,7 @@ class Command(BaseCommand):
         is_staff = options.get("is_staff")
 
         user_data = {}
-        user_password = ""
+        user_password = options.get("password")
         group = ""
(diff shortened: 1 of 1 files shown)


cdubz:
Looks good -- thank you!

PR #566, opened 2022-11-01
Update french translation / modification def dayssince in duration.py

Anoerak:
Update of the french translation (corrections and adds) + modification of the 
'def dayssince(value, today=None)' by replacing the value of line 118 with a variable in line 110 which will make the tranlsation easier for the future.


Anoerak:
I think I know why the def dayssince test failed.
Since I changed the def and replace the return with a variable build as 
`
days_ago = _(" ") + str(delta.days) + _(" days ago")
`
in order to ease the translation (in french, we would replace the " " by "il y a " and the " days ago" by " jours"), I needed to modify the expected results of the test from "x days ago" to " x days ago".
Should I close this PR and submit a new one with the corrected test?

cdubz:
@Anoerak I think the approach we should take here instead is to bring the `str(delta_days)` _in to_ the translatable string -- then you'll be able to put whatever you want before it without the filler space.

E.g. `_("%(days_ago)s days ago") % {"days_ago": str(delta.days)}`

Anoerak:
Ok.
I'll look into that this weekend.

cdubz:
@Anoerak 👋 Still looking to contribute to this branch?

Anoerak:
Hi @cdubz ,
Sorry, things have been busy here :)
I've to meet some deadlines on a current project and will jump back on this branch right after if that's OK with you.

cdubz:
Yep that’s fine. No worries!

cdubz:
Thanks for you work here, @Anoerak! I only took this over because of the related #718 change.

PR #577, opened 2022-11-24
Allow setting `CSRF_COOKIE_SECURE` and `SESSION_COOKIE_SECURE` via env vars

tpansino:
Added support and docs for setting `CSRF_COOKIE_SECURE` and `SESSION_COOKIE_SECURE` via env vars.

(It was possible to do this before with a custom settings module, but getting that configured, especially when running via Docker, is a lot of work compared to just setting an environment variable.)


cdubz (approved):
Looks good. Thanks for the PR!

PR #585, opened 2023-01-04
Add temperature to reports

pgazda:
Add report for temperature readings.
I basically just copied existing weight report. Only change worth mentioning is graph time range selectors. My reasoning is that parent is probably interested in temperature changes over past couple of hours or days (when the child might be sick) instead of past couple of weeks. That's why I used 12h, 24h, 48h, 3d and 7d.

### DIFF
--- a/reports/graphs/temperature_change.py
+++ b/reports/graphs/temperature_change.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+from django.utils.translation import gettext as _
+
+import plotly.offline as plotly
+import plotly.graph_objs as go
+
+from reports import utils
+
+
+def temperature_change(objects):
+    """
+    Create a graph showing temperature over time.
+    :param objects: a QuerySet of Temperature instances.
+    :returns: a tuple of the graph's html and javascript.
+    """
+    objects = objects.order_by("-time")
+
+    trace = go.Scatter(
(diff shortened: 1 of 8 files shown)


reports/graphs/temperature_change.py
```diff
@@ -0,0 +1,34 @@
...
+    :param objects: a QuerySet of Temperature instances.
+    :returns: a tuple of the graph's html and javascript.
+    """
+    objects = objects.order_by("-time")
+
+    trace = go.Scatter(
+        name=_("Temperature"),
+        x=list(objects.values_list("time", flat=True)),
+        y=list(objects.values_list("temperature", flat=True)),
+        fill="tozeroy",
```
cdubz, line 22: One minor thing here -- lets not use `fill` for this graph in particular. I think the presentation is kind of clearer without it (i.e., it makes it easier to see drastic changes or trends for the temperature).

cdubz (changes requested):
Looks good -- I have just one small CR to remove the fill for the graph. Thanks!

reports/graphs/temperature_change.py
pgazda: Yeah that makes sense.

cdubz (approved):
Looks good -- thanks!

Issue #587, opened 2023-01-16
Weight Change in Dashboard is Incorrect


tango2590:
Our newborn only has two weights in so far, both in grams. Initial weight was 2490g, the second weight is 2420g. It's only a 70g change, but the Statistics box on the Dashboard shows a weight change of -490.

<img width="536" alt="Screen Shot 2023-01-16 at 03 43 07" src="https://user-images.githubusercontent.com/78666662/212647207-21438af8-db3c-4eda-b01c-7c1994a77a62.png">
<img width="1523" alt="Screen Shot 2023-01-16 at 03 42 53" src="https://user-images.githubusercontent.com/78666662/212647212-474d3909-eb11-45e3-8dd2-278a53bfbb43.png">

cdubz:
Haha yeah that is weird... it's essentially projecting as if the trend will continue over the course of a full week. Probably this statistic just should not be shown if there is less than one week of data.

The relevant code is in `babybuddy/dashboard/templatetags/cards.py`:

https://github.com/babybuddy/babybuddy/blob/2690ab487614a4bf1a8f945ba87f47bb9a9c605c/dashboard/templatetags/cards.py#L585-L605

Thanks for reporting!

LoganBrinsmead:
I think I would like to take a try with this issue.

cdubz:
Yeah that's the bug -- once there is seven or more days between two entries it will just report the actual average.

rom1dep:
> it will just report the actual average.

…and just to be clear, that's precisely what I have a problem with. For several reasons:

- Reporting projections as facts may discourage from doing regular checks/measurements, giving a false sense of confidence and control
- This can in turn hide developmental issues that this app stands to detect and avoid
- Even if reporting projections as facts was OK, using averages in this context is painfully wrong: growth is non-linear, and a correct estimate "under ideal circumstances" (well fed, well rested, no medical condition, …) would require navigating [growth charts](https://www.cdc.gov/growthcharts/), which I'm convinced is a good addition to babybuddy, but is a big can of worms on its own (I'm not aware that they exist as open data, they are country-dependent and are revised over time, etc).

IMO this (and other similar) box(es) should contain something like: "Weight change: -70g over 2 days (4 days ago)"

cdubz:
> * This can in turn hide developmental issues that this app stands to detect and avoid
> 
[...]

PR #589, opened 2023-01-18
[RFC] proposal for "non-deceptive" stats

rom1dep:
Proposal for https://github.com/babybuddy/babybuddy/issues/587

### DIFF
--- a/dashboard/templatetags/cards.py
+++ b/dashboard/templatetags/cards.py
@@ -387,8 +387,9 @@ def card_statistics(context, child):
         stats.append(
             {
                 "type": "float",
-                "stat": weight["change_weekly"],
-                "title": _("Weight change per week"),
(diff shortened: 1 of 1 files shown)


cdubz:
@rom1dep I do think this change looks good -- can you fix the linting issues and push?

rom1dep:
Sure, but shouldn't this approach be generalized to other stats first? If so, I'll need to create the time in between bottles and diapers for it :)

cdubz:
Yep! Let’s do that here as well.

cdubz:
@rom1dep are you still interested in working on this branch?

rom1dep:
@cdubz Hi!
Interested? Sure!  
But realistically I don't expect to have time for that in the short, unfortunately.

Issue #584, opened 2022-12-19
Pumping entries should be taggable


btoconnor:
It would be good if pumping entries had tags.  We have more than one pump and my wife would like to associate which entry comes from which pump - tags seems like a reasonable solution to this (rather than her manually entering things).  It seems like a fairly straightforward change to accomplish as well (based on a quick look at the code - if that's not the case would definitely be open to correction!)

cdubz:
Yeah this an oversight. Everything should be taggable and I thought that was already the case 😃

cdubz:
Closed by #590

PR #590, opened 2023-01-28
Add tags to Pumping model

btoconnor:
Issue: #584 

This adds tags to the Pumping model.

Note: As also mentioned in #576 - I am having troubles with the test suite running properly. I got a bunch of failures unrelated to the Pumping model.  I might try recreating the environment, but if there's any guidance there it would be appreciated.  Apologies if the test suite fails on this - it was hard to get signal on it - hoping the Github checks will give some insight.


cdubz (changes requested):
@btoconnor this looks good for getting the tag relationship on there. We'll also want to add the tags to the pumping list view and filters. You can see examples of this in the other tag-supported models.

cdubz:
Also FTR I have manually deployed this to http://pr-590.dokku.baby-buddy.net/ -- still haven't had time to try to work out this damn GH action...

btoconnor:
I knew I missed something :)

Will get that going shortly.  Thanks for the tip.

cdubz:
Oh and the API, too! Check out the `TaggableSerializer` and implementation for other models.

cdubz:
Now it looks like we have consistent test fails. That's a good thing! I wonder if we have this on the main branch, too... 👀

cdubz:
Yeah there are inconsistent test fails on the main branch as well. I'll try to get things cleaned up there.

cdubz:
And now we're passing again. Something is definitely wonky here... anyway nothing this PR is doing is causing any fails (:

cdubz (changes requested):
@btoconnor thank you! One last thing -- to implement the tag filtering in the list view `core.filters.PumpingFilter` needs one small change. See the other filters for an example.

btoconnor:
Aha! I didn't even know about that feature - good to know!

cdubz (approved):
Looks good. Thanks!

btoconnor:
@cdubz Perfect - thanks for the assistance with getting this over the finish line.

Definitely no rush on this - this is pure curiosity - but when do these releases typically make it over to lscr.io for the Docker images?

cdubz:
I think only when I tag releases. I should be more active than I have been lately for the next week or so and I’m surely I’ll do at least one release in that time 😀

Issue #588, opened 2023-01-17
Implement ordering support in API


cdubz:
The API does not have support for ordering. DRF provides seemingly straightforward [ordering support](https://www.django-rest-framework.org/api-guide/filtering/#orderingfilter) so we should implement it.

PR #598, opened 2023-02-10
Add DRF's OrderingFilter to list of default filter backends to implement ordering in the API

btoconnor:
Issue: #588 

This adds DRF's `OrderingFilter` to the default list of filter backends, which allows us to include an `ordering` parameter in API requests.  AFAICT in the docs, this allows us to order on any fields that are in the serializer.

If we want to restrict that further, it's possible, but this felt like a good start.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -16,6 +16,8 @@ class BMIViewSet(viewsets.ModelViewSet):
     queryset = models.BMI.objects.all()
     serializer_class = serializers.BMISerializer
     filterset_fields = ("child", "date")
+    ordering_fields = ("child", "date")
+    ordering = "-date"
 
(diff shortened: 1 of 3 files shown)


cdubz:
Thanks for starting this out @btoconnor -- I added some of the view-level metadata for ordering as well.

btoconnor:
So - I had started doing a bunch of this but that's a little what I was alluding to with my comment - if these fields aren't specified, any field in the serializer can be ordered on, which felt mostly good enough for me? But happy, obviously, to be more specific.

Although I guess I should have gone with the defaults!

cdubz:
Yeah it could’ve stayed as-is but I like specificity 😃

PR #609, opened 2023-03-10
Add django-specific notes for PROXY_HEADER config

### LINKED ISSUE #608: PROXY_HEADER - Documentation update
This isn't really an issue with your docs exactly, and you do specify to read the the django docs [here](https://docs.djangoproject.com/en/4.1/howto/auth-remote-user/#configuration).

When setting up this config -- I found that you had to convert all headers to follow the django spec for [request.meta](https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.HttpRequest.META)  namely this part:

> With the exception of CONTENT_LENGTH and CONTENT_TYPE, as given above, **any HTTP headers in the request are converted to META keys by converting all characters to uppercase, replacing any hyphens with underscores and adding an HTTP_ prefix to the name. So, for example, a header called X-Bender would be mapped to the META key HTTP_X_BENDER.**

As a result, using a header (like the one you provided in your example) as `X-Auth-User` would be have to be converted to `HTTP_X_AUTH_USER`

If you could add a line similar to what's included in the docs for how headers are mapped, that would go a long way in reducing the time it takes to figure this out.

p-b--:
Fixes #608

### DIFF
--- a/docs/configuration/security.md
+++ b/docs/configuration/security.md
@@ -54,13 +54,16 @@ Each entry must contain both the scheme (http, https) and fully-qualified domain
 Sets the header to read the authenticated username from when
 `REVERSE_PROXY_AUTH` has been enabled.
 
+Django modifies headers in the HTTP request; HTTP headers in the request have all characters converted to uppercase, replacing any hyphens with underscores and adding an HTTP_ prefix to the name. For example `X-Auth-User` would be converted to `HTTP_X_AUTH_USER`.
+
 **Example value**
 
+    // For header key X-Auth-User
     HTTP_X_AUTH_USER
 
 **See also**
-
 - [Django's documentation on the `REMOTE_USER` authentication method](https://docs.djangoproject.com/en/4.1/howto/auth-remote-user/)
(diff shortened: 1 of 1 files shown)


docs/configuration/security.md
```diff
@@ -54,13 +54,16 @@ Each entry must contain both the scheme (http, https) and fully-qualified domain
 Sets the header to read the authenticated username from when
 `REVERSE_PROXY_AUTH` has been enabled.
 
+Django modifies headers in the HTTP request; HTTP headers in the request have all characters converted to uppercase, replacing any hyphens with underscores and adding an HTTP_ prefix to the name. For example `X-Auth-User` would be converted to `HTTP_X_AUTH_USER`.
```
cdubz, line 57: One very tiny CR here -- while Django is referenced throughout the links and indeed the thing doing this I try to make the docs less confusing by (mostly) avoiding direct references to technical dependencies. Lets just change this to "Baby Buddy modifies [...]".

cdubz (changes requested):

docs/configuration/security.md
p-b--: Sure -- done.

cdubz (approved):
Looks good, thanks!

PR #627, opened 2023-04-15
v2

### LINKED ISSUE #515: Release version 2.0.0
I have started work recently on a [v2 branch](https://github.com/babybuddy/babybuddy/tree/v2) and will try to focus future efforts on moving that forward with some issues with breaking changes. I'll use this issue on tracking for the relevant issues that need to be completed.

## Planned release date

26 May

### Breaking changes

- `TIME_ZONE` variable no longer supported. User timezone can be set from user settings.
- `NAP_START_MIN` and `NAP_START_MAX` variables no longer supported. Nap time parameters can be set from site settings.
- `Timer` model `end` and `duration` fields removed. There is no longer a concept of "active" vs. "inactive" timers.
- API: `active` filter on `Timer` model no longer supported.
- API: `PATCH /api/timers/$x/stop` endpoint no longer supported.

### Major changes

- Custom datetime selector has been removed. Browser defaults for locale datetime fields are now used.
- Timers no longer have an "end" and cannot be stopped -- only deleted or used to create a new entry with times.
- There is a new site-wide settings option under user menu.
- Sleep entries can now be defined as a "nap" (or not) manually.
- 24-hour time override has been removed. Time format will be determined by browser datetime fields based on locale.

## v2.0.0

- [x] #635
- [x] #301
- [x] #636
- [x] #637
- [x] #638
- [x] #109 
- [x] #639
- [x] #513 

## Stretch goals

- [x] #470 
- [ ] #471
- [ ] #460
- [ ] #161 
- [ ] #34

cdubz:
TODO:

- [x] Fix datetime field issues for Safari
- [x] Override site settings form
- [x] Add site settings tests
- [x] Update localization strings


cdubz:
Prefilled fields from timers are not always recognized as valid. This example is in Firefox on iOS (Safari) —

image

The issue can worked around by activating each field and selecting the time. Seems to be some problem with how the default values are populated. 

Default values _do_ work when tested e.g. in Firefox on Ubuntu.

PR #647, opened 2023-05-16
Feature/added hungarian translation

ssilverr:
Added Hungarian translation. 

Having trouble with setting up the project locally though. Lint fails on the migration files :/


ssilverr:
Oh, I guess I missed the fact that a Hungarian translation was added in the meantime

Issue #573, opened 2022-11-13
Pumping should have a start time and be creatable from a timer


hsharrison:
Just setting this up and looks like it meets pretty much all our needs (for the last child we used an Android app which was good but didn't sync between us), this is one thing suggested by my wife. Let's see if I can find time to look into it myself, but thought I may as well open an issue.

cdubz:
Makes sense! Will watch out for a PR. Shouldn’t be too complicated.

tzw9373:
I don't have the skills to assist with a PR, but just wanted to say this feature was requested by my wife as well and would be greatly appreciated!

moejay:
I'll give this a shot, my wife loves the app, and requested the exact same feature.

johnmerlie:
I've begun to look at this, as my wife had the same request. I've started working on this at [johnmerlie:babybuddy](https://github.com/johnmerlie/babybuddy/tree/pumping-end-time). 

I think i've got most of the features down, as it was largely a process of mimicking the setup of the `feeding` entity.

I've tested in-situ (via `gulp`) by adding pumping entries via the list view as well as the quick timer - but as this is my first experience with django migrations I'm not sure if i've completely addressed that from the standpoint of modifying the existing schema. The strategy I'm thinking is to simply convert the `pumping.time` field to a `pumping.start` and add a `pumping.end` field - populating the existing `pumping.start` time in that field to ensure duration can be calculated but ultimately showing the record as a 0 duration pumping entry. 

I'll also still need to look at the tests, as I know there will be some additional work needed there.

PR #665, opened 2023-06-16
Add Pumping Duration and Create from Quick Timer

johnmerlie:
This PR covers adding start/end times and duration (as exist on `feeding`) to `pumping`, as well as adding the ability to create a `pumping` entry from a quick timer as discussed in #573


cdubz:
@johnmerlie this is all looking about right. I'd like to get the _existing_ tests updated before I do a thorough review here.

It looks like a big chunk of the current fails are just related to the now outdated fixtures. There is only one pumping entry in the fixtures:

https://github.com/johnmerlie/babybuddy/blob/1fef726b9b127ed7349bbcba40987da6e420ef29/babybuddy/fixtures/tests.json#L24-L34

You should be able to essentially update that by hand and then get a better idea of what tests actually do need to be updated.

johnmerlie:
@cdubz Thanks for the guidance. I've gone back and updated all the test data/cases, including adding an additional pumping entry (to double check sorting changes). 

Looks like everything is passing locally when I run gulp test, and checks are passing. 

The one thing I wasn't sure about is whether there's a good way to test the migration on existing data - from what i can tell it looks like all the fake data for testing is generated _after_ the migration command, so it's naive to old models. If that's expected and explicit testing of the data migrations isn't explicitly tested, then this should be about ready.

Let me know what you think

core/migrations/0029_alter_pumping_options_remove_pumping_time_and_more.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 4.2.2 on 2023-06-16 20:06
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+def set_pumping_start_end_values(apps, schema_editor):
+    # The model must be imported to ensure its overridden `save` method is run.
+    from core.models import Pumping
```
cdubz, line 9: To ensure you have access to the old `time` field you'll need to get `Pumping` from `apps` -- e.g., `Pumping = apps.get_model("core", "Pumping")`. The `apps` variable represents the model _before_ the migration.

api/views.py
```diff
@@ -82,8 +82,8 @@ class PumpingViewSet(viewsets.ModelViewSet):
     queryset = models.Pumping.objects.all()
     serializer_class = serializers.PumpingSerializer
     filterset_class = filters.PumpingFilter
-    ordering_fields = ("amount", "time")
-    ordering = "-time"
+    ordering_fields = ("amount", "duration", "end", "start")
+    ordering = "-start"
```
cdubz, line 86: For now lets follow the same default ordering as `Feeding` and set this to `-end`.

core/templates/core/pumping_list.html
cdubz, line 1: Rename the column header here `Date` just to keep this in parity with the Feeding list.

cdubz (changes requested):
Mostly looks good! Just a few small CRs here.

> The one thing I wasn't sure about is whether there's a good way to test the migration on existing data - from what i can tell it looks like all the fake data for testing is generated after the migration command, so it's naive to old models. If that's expected and explicit testing of the data migrations isn't explicitly tested, then this should be about ready.

I have never got around to building in any sort of migration testing. I probably _should_ but for now it's just manual. So no worries on that front.

core/migrations/0029_alter_pumping_options_remove_pumping_time_and_more.py
johnmerlie: fixed with [b7b28926c164a6cae38ff750fdbde79f1b2fa3e6](https://github.com/babybuddy/babybuddy/pull/665/commits/b7b28926c164a6cae38ff750fdbde79f1b2fa3e6)

api/views.py
johnmerlie: fixed with [b8a1e39987cfcad5a7d59cb60b0e153812dc2b94](https://github.com/babybuddy/babybuddy/pull/665/commits/b8a1e39987cfcad5a7d59cb60b0e153812dc2b94)

core/templates/core/pumping_list.html
johnmerlie: fix for [2485f0309225f61f04058c1f2c7b19267c9bd000](https://github.com/babybuddy/babybuddy/pull/665/commits/2485f0309225f61f04058c1f2c7b19267c9bd000)

johnmerlie:
@cdubz  Should all be fixed, I'll let you take a look and mark as resolved if you agree

> I have never got around to building in any sort of migration testing. I probably should but for now it's just manual. So no worries on that front.

No worries, just wanted to make sure I wasn't missing anything!

cdubz (approved):
Looks good. Thanks for all your work on this!!

johnmerlie:
> Looks good. Thanks for all your work on this!!

@cdubz babybuddy has been a huge help with our (almost) 7 week old, so it feels like the least I can do!

PR #672, opened 2023-06-25
Use homeassistant X-Ingress-Path for building paths, expose ingress_session

MrApplejuice:
Relates to https://github.com/OttPeterR/addon-babybuddy/issues/47

## Changes

- Add a middleware that exposes the X-Ingress-Path via the Django-Http request object for later use
- Added settings-option to allow the middleware to use the X-Ingress-Path value to construct abolsute paths
    - This fixes a bunch of things in home assistant, for example, that paths generated by django rest framework are correct now! :partying_face: 
    - This fixes the absolute login path generated for and encoded in the QR-code when everything is run through home assistant
- The option to use X-Ingress-Path defaults to "OFF", added a special homeassistant config that enables the option
- In addition, the qr-code generating code searches and exposes the ìngress_session`cookie created and exposed by homeassistant. This allows the android app to use the same session to login to babybuddy hosted via the ingress-service of home assistant. _I am not quite sure if this is the correct way of doing this, I will try it out myself a bit to see if this works or not._

### DIFF
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -98,9 +98,9 @@ class HomeAssistant:
 
     def __call__(self, request: HttpRequest):
         def wrap_x_ingress_path(org_func):
-            if request.headers.get("HTTP_X_HASS_SOURCE") != "core.ingress":
+            if not request.is_homeassistant_ingress_request:
                 return org_func
-            x_ingress_path = request.headers.get("HTTP_X_INGRESS_PATH")
+            x_ingress_path = request.headers.get("X-Ingress-Path")
             if x_ingress_path is None:
                 return org_func
 
@@ -114,6 +114,10 @@ class HomeAssistant:
 
                 return url
             return wrapper
+
+        request.is_homeassistant_ingress_request = (
+            request.headers.get("X-Hass-Source") == "core.ingress"
+        )
 
         if self.use_x_ingress_path_rewrite:
             request.build_absolute_uri = wrap_x_ingress_path(
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -10,6 +10,7 @@ from django.utils import timezone, translation
 from django.contrib.auth.middleware import RemoteUserMiddleware
 from django.http import HttpRequest
 
+
 class UserLanguageMiddleware:
     """
     Customizes settings based on user language setting.
@@ -97,17 +98,17 @@ class HomeAssistant:
     Django middleware that adds HomeAssistant specific properties and checks
     to the request-object.
 
-    The middleware is only active if the settings variable 
+    The middleware is only active if the settings variable
     `ENABLE_HOME_ASSISTANT_SUPPORT` is set to True. Note that some features
-    remain enabled even if the middleware is set to inactive through the 
+    remain enabled even if the middleware is set to inactive through the
     settings.
 
     Features:
-    
+
     - request.is_homeassistant_ingress_request (bool)
 
         Indicates if a request was rerouted through the home assistant ingress
-        service. This parameters is always present regardless of the 
+        service. This parameters is always present regardless of the
         ENABLE_HOME_ASSISTANT_SUPPORT settings option. It defaults to false
         if the middleware is disabled.
 
@@ -139,11 +140,10 @@ class HomeAssistant:
                     url_parts._replace(path=x_ingress_path + url_parts.path)
                 )
                 return url
+
             return wrapper
 
-        request.build_absolute_uri = wrap_x_ingress_path(
-            request.build_absolute_uri
-        )
+        request.build_absolute_uri = wrap_x_ingress_path(request.build_absolute_uri)
 
     def __call__(self, request: HttpRequest):
         if self.home_assistant_support_enabled:
@@ -156,4 +156,3 @@ class HomeAssistant:
         self.__wrap_build_absolute_uri(request)
 
         return self.get_response(request)
-    
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -9,6 +9,7 @@ from django.utils import timezone, translation
 from django.contrib.auth.middleware import RemoteUserMiddleware
 from django.http import HttpRequest
 
+
 class UserLanguageMiddleware:
     """
     Customizes settings based on user language setting.
@@ -113,6 +114,7 @@ class HomeAssistant:
                 )
 
                 return url
+
             return wrapper
 
         request.is_homeassistant_ingress_request = (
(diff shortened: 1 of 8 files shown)


babybuddy/middleware.py
```diff
@@ -88,3 +90,38 @@ class CustomRemoteUser(RemoteUserMiddleware):
     """
 
     header = getenv("PROXY_HEADER", "HTTP_REMOTE_USER")
+
+
+class HomeAssistant:
```
cdubz, line 95: I'd like to see more in code documentation here (and throughout the class) explaining what this is doing and _why_. I've spent a bit of time reviewing it and I *think* I understand based on various (weak) sources of information from HA forms and the like. Future me (and anyone else contributing code) would be thankful for more direct documentation here.

babybuddy/settings/base.py
```diff
@@ -351,9 +352,13 @@
...
+# See https://docs.baby-buddy.net/ for details about these settings.
 
 BABY_BUDDY = {
     "ALLOW_UPLOADS": bool(strtobool(os.environ.get("ALLOW_UPLOADS") or "True")),
     "READ_ONLY_GROUP_NAME": "read_only",
 }
+
+# Home assistant specific configuration
+
+HOME_ASSISTANT_USE_X_INGRESS_PATH = False
```
cdubz, line 364: We'll want to expose this via an env variable. See e.g. `SESSION_COOKIE_SECURE` for an example.

babybuddy/views.py
```diff
@@ -251,7 +253,15 @@ class UserAddDevice(LoginRequiredMixin, View):
     qr_code_template = "babybuddy/login_qr_code.txt"
 
     def get(self, request):
-        qr_code_response = render(request, self.qr_code_template)
+        session_cookies = {}
```
cdubz, line 256: Lets also get some documentation here about what this is doing and why.

Also since we are unsure of this approach... are there any security concerns here?

babybuddy/middleware.py
```diff
@@ -88,3 +90,38 @@ class CustomRemoteUser(RemoteUserMiddleware):
...
+
+                return url
+
+            return wrapper
+
+        request.is_homeassistant_ingress_request = (
+            request.headers.get("X-Hass-Source") == "core.ingress"
+        )
+
+        if self.use_x_ingress_path_rewrite:
```
cdubz, line 124: Is there any reason one would _not_ want this functionality if we know this is an ingress request? I.e., could we not just rely on the `request.is_homeassistant_ingress_request` instead of a requiring a bespoke configuration toggle?

mkdocs.yml
```diff
@@ -19,6 +19,7 @@ nav:
     - 'configuration/email.md'
     - 'configuration/security.md'
     - 'configuration/storage.md'
+    - 'configuration/homeassistant.md'
```
cdubz, line 22: Move this in to alpha order.

cdubz (changes requested):
Aside from inline CRs I'd like to see a test added for this as well. Once we have the approach down.

babybuddy/views.py
MrApplejuice: > Documentation

:+1:

> Also since we are unsure of this approach... are there any security concerns here?

Good question. And honest answer: I am not sure. I think the `ingress_session` is specific to the ingress-service of home assistant. If someone uses it, the session needs to come into existence somehow and something needs to be cached. This is certainly not the user's password. It might even be the case that this is a very specific key that is specific to the home assistant app, which I think would be the desired approach.

Unfortunately, the home assistant documentation of the ingress service seems to be quite "thin" and the codebase seems to be very very opaque.

Personal opinion: Since the ingress service requires a valid home assistant user session, represented by the ingress_session as far as I can tell, this seems to be a requirement for accessing home assistant in any case. It is not the password, it is specific to the ingress service. I think this is probably ok... worst thing that I experienced is that those session tokens timeout and become unusable. There might be a better way using the persistent API-tokens for home assistant, but those give access to more home assistant features...

Conclusion: :shrug: I am not quite sure, I think this is "ok" but maybe needs to be replaced or complemented in the future...

babybuddy/settings/base.py
MrApplejuice: Little clarification: I now use a custom `homeassistant.py` config with these contents:

~~~~~
from .base import *

HOME_ASSISTANT_USE_X_INGRESS_PATH = True
~~~~~

to do the same thing. You would prefer env-vars for this? I could also add the custom settings-file alternatively? Or do you prefer envvars generally?

cdubz: I mostly rely on env vars to ensure Docker users are supported without having to fiddle with file management and storage stuff. If HA is not via Docker (is it, I thought it was?) I suppose it's OK to keep this without an env var.

babybuddy/views.py
cdubz: > Unfortunately, the home assistant documentation of the ingress service seems to be quite "thin" and the codebase seems to be very very opaque.

Yeah I'm not able to find much on this, either. I'm wondering about session hijacking here but lacking any clear information or documentation it's hard to judge. I'm OK to leave it.

babybuddy/middleware.py
cdubz: Ok, this is all makes sense.

> ... now that I wrote this down however, I realize that this is not quite what I implemented. The check if a request is an ingress request is happening regardless of the "active state" of the middle ware. I will fix that one. That one should also be

Yeah I think this is part of where I was confused 😄

babybuddy/settings/base.py
MrApplejuice: Nah, it is docker based. So an env-var party it is!

babybuddy/views.py
MrApplejuice: Coming back to the documentation mainly:

I added some, but I feel rather than adding docs to `babybuddy/views.py` a doc-page might be even more suited. Right now, there is the qr code, but there is no information anywhere how a third-party developer can use it to authenticate with babybuddy. So, maybe a new doc-category aimed at app developers might be more fitting, which describes the json-structure in the qr-code and how to use it?

babybuddy/middleware.py
MrApplejuice: Fixed!

MrApplejuice:
Okay, I think I addressed all the open issues now. One more question about the qr-code docs (see above).

I also found some time to add some tests for the x-ingress-path stuff. Thought that might be handy.

babybuddy/views.py
cdubz: That makes sense but deserves a separate issue, I think. I have created #674 for it.

cdubz (approved):
All looks good! Thanks for adding the test as well!

PR #700, opened 2023-08-15
[Feature] Add Filter timers by name

bstaeheli:
This commit introduces a new feature that allows users to filter timers by their names. With this enhancement, users can now easily locate timers even if they don't know their exact IDs or are unaware of the existence of certain timers. This improvement facilitates seamless integration with programming scenarios like Apple Shortcuts, enabling more elegant solutions.


cdubz (approved):
Makes sense. Thanks for contributing!

PR #707, opened 2023-09-17
feat: devenv.sh for a reproducible developer environment (built on nix)

barakplasma:
my version of PR #549 
uses devenv.sh for a reproducible developer environment (built on nix).

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+source_url "https://raw.githubusercontent.com/cachix/devenv/d1f7b48e35e6dee421cfd0f51481d17f77586997/direnvrc" "sha256-YBzqskFZxmNb3kYVoKD9ZixoPXJh1C9ZvTLGFRkauZ0="
+
+use devenv
--- a/README.md
+++ b/README.md
@@ -7,9 +7,10 @@
 [![CI Status](https://github.com/babybuddy/babybuddy/actions/workflows/ci.yml/badge.svg)](https://github.com/babybuddy/babybuddy/actions/workflows/ci.yml)
 [![Coverage Status](https://coveralls.io/repos/github/babybuddy/babybuddy/badge.svg?branch=master)](https://coveralls.io/github/babybuddy/babybuddy?branch=master)
 [![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/babybuddy/babybuddy)
+[![Open in GitHub Codespaces ready-to-code](https://img.shields.io/badge/Codespace-ready--to--code-blue?logo=github)](https://codespaces.new/babybuddy/babybuddy?quickstart=1)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 
-A buddy for babies! Helps caregivers track sleep, feedings, diaper changes, 
+A buddy for babies! Helps caregivers track sleep, feedings, diaper changes,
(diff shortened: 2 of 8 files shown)


cdubz:
Interesting. Thanks for sharing. Would this be an alternative to #529?

barakplasma:
Yes, would be an alternative to #549 .

Also enables GitHub codespaces I think, as an alternative to gitpod

devenv.lock
cdubz, line 1: Should the lock file be committed? Will it work for other architectures? A while back I stopped committed the pipenv lockfile because it the lock only applied to a single architecture. I don't know if thats an issue for devenv (or even if thats still the case for pipenv... this was a long time ago).

.envrc
cdubz, line 1: What is this source doing? It seems to be some kind of devenv configuration? I don't love referencing raw files from GH...

And what uses this `.envrc` file?

README.md
```diff
@@ -7,9 +7,10 @@
 [![CI Status](https://github.com/babybuddy/babybuddy/actions/workflows/ci.yml/badge.svg)](https://github.com/babybuddy/babybuddy/actions/workflows/ci.yml)
 [![Coverage Status](https://coveralls.io/repos/github/babybuddy/babybuddy/badge.svg?branch=master)](https://coveralls.io/github/babybuddy/babybuddy?branch=master)
 [![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-blue?logo=gitpod)](https://gitpod.io/#https://github.com/babybuddy/babybuddy)
```
cdubz, line 9: Would any of this stuff be advantageous for Gitpod as well? I don't have a strong opinion about including this... just curious if there is an easy improvement here as well.

docs/contributing/development-environment.md
```diff
@@ -76,3 +76,13 @@ information and steps below to set up a local development environment for Baby B
 
 Open [http://127.0.0.1:8000](http://127.0.0.1:8000) and log in with the default
 username and password (`admin`/`admin`).
+
+## Alternative: https://devenv.sh/
```
cdubz, line 80: This is actually smooth enough that I'm inclined to make it the primary development environment method... is devenv the only direct dependency?

cdubz (changes requested):
This is really wonderful! And so easy.

Before merging I have a few questions posted via review... and maybe some small changes needed.

Thanks, again, for this work!

devenv.lock
barakplasma: I'm not sure about the arch of the lockfile, but I made this on amd64 and I have an arm server so I can test it out there.

The lockfile should be committed since it has the exact versions of all the dependencies so the devenv is reproducible

.envrc
barakplasma: The envrc is a devenv thing; it's like the distro version I think? I'll dig a little and let you know

README.md
barakplasma: I like gitpod, but I lost track of it after GitHub codespaces came out. We could maybe get gitpod to use the devenv config if you want me to look into it. I think they are competitors to be honest.

docs/contributing/development-environment.md
barakplasma: That's the beauty of devenv: it's the only dependency you need to install since it bootstraps everything else. Docker isn't a requirement. 

However, on windows, WSL2 or docker is a requirement. I actually made this PR with WSL2, but I also tested the vscode devcontainer, and GitHub codespaces, and they work well too.

cdubz (approved):
Ok lets give this a try. I think we may want to tweak things as we go but this seems like a good foundation. Thanks 👍

PR #444, opened 2022-04-16
New statistic graphs: Diaper change intervals & feeding intervals

myxor:
New statistic graphs: Diaper change intervals & feeding intervals

grafik

grafik

PR #710, opened 2023-09-30
feat: statistics on diaper change intervals

barakplasma:
Continuation of #444 but with the git history sorted out.
Includes the fix from @garyglatfelter for graph starting from left axis


cdubz (approved):
Looks good -- thanks for getting this fixed up!

PR #711, opened 2023-09-30
feat: configure and explain debugging in devenv

barakplasma:
while testing PR 710 I wanted to debug it so I setup debugging support

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -18,8 +18,11 @@
     pipenv install --dev
     npm i
     gulp migrate
-    gulp
   '';
+
+  processes = {
+    gulp.exec = "gulp";
+  };
 
   # https://devenv.sh/languages/
   languages.python.enable = true;
@@ -31,6 +34,7 @@
   devcontainer = {
     enable = true;
     settings.customizations.vscode.settings."[python]".editor.defaultFormatter = "ms-python.black-formatter";
+    settings.customizations.vscode.settings."[python]".pipenvPath = "pipenv";
     settings.customizations.vscode.extensions = [
       "bbenoist.Nix"
       "ms-python.python"
--- a/docs/contributing/development-environment.md
+++ b/docs/contributing/development-environment.md
@@ -81,8 +81,41 @@ username and password (`admin`/`admin`).
 
 1. Install devenv (linux/mac/windows subsystem for linux/docker) via https://devenv.sh/getting-started/
 1. git clone the repo
-1. run `devenv shell`
+1. here you have two options
+    1. run `devenv shell` and then run `gulp`
+    1. _OR_ run `devenv up` which automatically runs gulp/nodejs/python as well
 1. Open [http://127.0.0.1:8000](http://127.0.0.1:8000) and log in with the default
 username and password (`admin`/`admin`).
 
-Devenv also generates the devcontainer for vscode, and sets up github codespaces support. To run babybuddy, open a terminal after opening this repo in the devcontainer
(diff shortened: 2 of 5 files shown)


.vscode/launch.json
barakplasma, line 1: we don't need to include this file for debugging to work; the copy in the docs would suffice.

devenv.nix
```diff
@@ -18,9 +18,12 @@
     pipenv install --dev
     npm i
     gulp migrate
-    gulp
```
barakplasma, line 21: This change is necessary for the vscode debugger to have a chance to start django itself (rather than via gulp). The caveat is the extra step of running `gulp` in devcontainers or `devenv up` instead (thus the change to the devenv docs)

cdubz (changes requested):
This awesome! Thank you! I come primarily from a PHP development background and was never really able to sort out breakpoint debugging in Python. This makes it so easy.

Two additional ones (one directly related, one not so much):

Directly related:

When I follow the debugger documentation steps in Codespaces all the static files respond 404 so I don't get any CSS or JS in the app. I assume this is probably somehow related to the `STATICFILES_STORAGE` setting but wonder if you can reproduce/have any thoughts before I look closer at it.

Not so related:

Whenever I open the code in Codespaces I get this notification while things are still starting up. Is this expected and/or is there anything we can do to prevent it? It's just a bit confusing.

Screenshot from 2023-09-30 06-41-37

.vscode/launch.json
barakplasma: Sounds good.
Including a template / just docs matches the convention in other OSS repos like:
- https://github.com/microsoft/TypeScript/blob/main/.vscode/launch.template.json
- https://github.com/electron/electron/blob/371e83a8d2390c63bc9171c0b3949b460ab9ad17/docs/tutorial/debugging-vscode.md?plain=1#L16-L36
- https://github.com/scrapy/scrapy/blob/master/docs/topics/debug.rst#visual-studio-code

done in 9a7f095a5ed39e074def238e3e7d4e2d7d6d02fd

docs/contributing/development-environment.md
```diff
@@ -81,8 +81,41 @@ username and password (`admin`/`admin`).
...
+            "args": [
+                "runserver"
+            ],
+            "django": true,
+            "justMyCode": true
+        }
+    ]
+}
+```
+2. **select the pipenv python interpreter** in vscode. The default python env won't have the dependencies needed for debug
```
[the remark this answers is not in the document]
barakplasma: Adding the following line to the devenv.nix file sets the interpreter in the codespace at least:

```nix
settings.customizations.vscode.settings."[python]".pipenvPath = "pipenv";
```

working for me in the codespace

barakplasma:
> Whenever I open the code in Codespaces I get this notification while things are still starting up. Is this expected and/or is there anything we can do to prevent it? It's just a bit confusing.

There's a newline added/removed occassionally from the devcontainer.json file when devenv shell runs that's causing this notification to happen in my experience. I am not sure what is causing it, or what the default should be, but I'm pretty sure that's the issue.

cdubz:
> When I follow the debugger documentation steps in Codespaces all the static files respond 404 so I don't get any CSS or JS in the app. I assume this is probably somehow related to the STATICFILES_STORAGE setting but wonder if you can reproduce/have any thoughts before I look closer at it.

Any thoughts on this piece? And do you get the same issue?

barakplasma:
> > When I follow the debugger documentation steps in Codespaces all the static files respond 404 so I don't get any CSS or JS in the app. I assume this is probably somehow related to the STATICFILES_STORAGE setting but wonder if you can reproduce/have any thoughts before I look closer at it.
> 
> Any thoughts on this piece? And do you get the same issue?

That happened to me because of the debug=true setting I had previously. I changed it to debug=false and now it doesn't happen

cdubz (approved):
Looks good!

PR #732, opened 2023-11-16
deployment baby buddy setup through GCP

NissesSenap:
Share a way to setup baby buddy in GCP using cloud run.


MrApplejuice:
Wow, wow, wow... this looks flipping awesome. You can scale the instances down to _zero_? That would save me so much money with the test-deployment server for the babybuddy app...

Can you comment on three questions?

- Do costs scale down to zero (or close to zero?) as well?
- Does the "offline server" maintain its IP address so that DNS does not need any updating?
- How long does it take (approximately) to go from zero to one again?

NissesSenap:
Yes, down to zero :)

* close to zero - the SQL cost which isn't super high.
* So this solution uses the generated URI that you get from GCP cloud run which got https and some other nice features. But if you want a custom DNS you will need to do some more stuff. So to answer your question, no the IP changes, and you need to do some more config to get your cloud run to use your current DNS record.
* Say that it takes 5 seconds with a cold start. One of the major issues is that the container image that we are using is 450 Mb big. I know it's best practice but the question is if you really need gunicorn to run this app. I assume most use cases are only 1-2 users at the same time and about 1-5 requests when the app is used. So gunicorn isn't really needed. If we just run with the django server we can probably save a few seconds in startup time. But then we will need to create a new container image.

So I have only run this for a single day, so I don't know, but I share the cost output so far. It's in currently 2,7 SEK, which translates to about 0.3 USD.
image
It's the SQL instance that cost money and sadly GCP don't have a serverless SQL solution or any free tier for the db.


I chose to use GCP due to their serverless solution, don't require any changes to the container.

AWS on the other hand got an RDS [free tier](https://aws.amazon.com/free/database/) for 12 months (you can always create a new account and get 12 new months) of their smallest RDS db. Another option is also to use their Aurora serverless DB [solution](https://aws.amazon.com/rds/aurora/serverless/). I couldn't find if any hours are included before you start getting charged.

AWS lambda is free in the [start](https://aws.amazon.com/lambda/pricing/) 
> The AWS Lambda free tier includes one million free requests per month and 400,000 GB-seconds of compute time per month

But to use lambda, we need to create a new container that is adjusted to AWS + do some minor code changes to respond to request the way lambda [wants](https://docs.aws.amazon.com/lambda/latest/dg/python-image.html#python-image-instructions).
Another option is of course to use the zip archive function, but we will still need to do some minor code changes. But I think we should be able to do them without creating any issues for the existing deployments. But I haven't looked in to it.

So in short, I think AWS looks better if you want to have the cheapest possible solution to host baby biddy. But I don't know AWS as well as GCP, so I went with what I know. But if you want, I can dust off the old AWS skills and see what I can come up with.

MrApplejuice:
Well, I did not know there is a possibility to have something spin up _that_ quickly with _that_ low costs in the first place. My digital ocean server is total overkill for this at the moment... 

<details>
  <summary>Here the costs and metrics for half of this month:</summary>

image
image

</details>

The main problem I am facing is the DNS. On digitalocean, I need a fixed IP to point my DNS there and that costs money. I will look into Google's offering I think to see if I can optimize this a bit. I am publishing on Android anyways. Makes sense to complete the circle... and also the server literally is not user-facing and only used for app-verfication.

Big thanks for the details and thank you for the work you put into this... I hope that I find some time to look into using your work!

cdubz:
Thanks for sharing this! I've been trying to find time to test it out but just having been able to get it 🚸

Based on @MrApplejuice's testing I'll merge this and we can tweak if necessary once I can do some more testing.

PR #740, opened 2023-12-17
Fix misc report breadcrumb issues

ntrecina:
### Fixed report breadcrumbs being duplicated.
The breadcrumbs throughout the report pages was duplicating the "Child Name / Reports" portion. This would happen regardless of whether you had one or more than one children saved.

Before
Screen Shot 2023-12-16 at 10 43 02 PM

After
Screen Shot 2023-12-16 at 10 43 45 PM

The breadcrumbs were also being duplicated on most reports which has also been fixed.

Before
Screen Shot 2023-12-16 at 10 46 03 PM

After
Screen Shot 2023-12-16 at 10 46 12 PM

### Fixed DiaperChangeTypesChildReport pointing to the wrong report view.
When going to the Diaper Change Types report it would instead go to the Diaper Change Intervals report.

Before
Screen Shot 2023-12-16 at 10 47 36 PM

After
Screen Shot 2023-12-16 at 10 47 48 PM

### Fixed child switcher for WHO weight percentiles boy/girl reports.
The "WHO Weight Percentiles for Boys/Girls in kg" reports would incorrectly redirect to the regular "Weight" report when switching between children via the quick switch dropdown in the breadcrumbs. I changed the WHO girl/boy reports to be a bit more independent of the regular weight report which enabled me to pass the independent urls to the child quick switch dropdown.


cdubz:
Thanks for catching and fixing these issues!

PR #743, opened 2023-12-22
742 fix sleep zh trans

hejhj:
Fix and add translation for the dashboard screen


cdubz:
Looks good -- thanks for contributing!

PR #752, opened 2024-01-21
742 fix sleep zh trans

hejhj:
add back zh translation fix


cdubz:
Thanks for fixing this (again!).

PR #756, opened 2024-01-26
feat: create a trimmed form just for bottle feeding.

billybonks:
While Feeding is great, I think there is room to create optimised forms for breastfeeding and bottle feeding.

This is the first one a trimmed-down form for bottle feeding specifically

image

### DIFF
--- a/babybuddy/templates/babybuddy/nav-dropdown.html
+++ b/babybuddy/templates/babybuddy/nav-dropdown.html
@@ -264,7 +264,12 @@
                                     {% trans "Feeding" %}
                                 </a>
                             {% endif %}
-
+                            {% if perms.core.add_feeding %}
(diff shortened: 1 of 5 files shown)


cdubz:
This seems like a good idea -- happy to review when it's ready 👍

billybonks:
Fixed the lint issues, i wanted to write a feature test

1. visit page
2. fill in input
3. submit
4. assert item in the list

But there are no facilities to do that. I saw view tests that do some basic HTTP testing. and form tests that assert the HTML provided by the view.

With that in mind, I think the best I could do is write a test to assert the title, but that doesn't seem to be high value.

Feel free to review; I will do the breastfeeding in a separate pr.

cdubz (approved):
Looks good! Thanks for contributing.

And yeah I have always wanted to add better features tests but just never got around to setting up. Makes sense not to add anything new here.

PR #760, opened 2024-02-06
Feat: Increase mobile friendliness of forms

billybonks:
While experimenting on https://github.com/babybuddy/babybuddy/pull/754. I explored variations of form layout with fewer input boxes and replaced drop downs for pills to reduce clicks on mobile.

What I don't want to do is create an entirely forked UI, so this PR attempted to add the elements but using the original skin.

I added the advanced option, as I believe that people will use tags and notes because there is something they cannot represent in the core data model, so ideally, it is not a core feature that people use. For that reason, it defaults to hidden until people use it. Then, it will use the last toggle state.

My implementation of the fieldset might not be ideal; I just did it so that I knew how to do it in Django (hardly code in Python/Django), so feel free to modify or let me know how you want it done.

Adding the toggle code for the details, as I did, is probably not ideal. I'm happy to change.

We could bring back the drop-downs on the desktop if we wanted to by rendering both and using a media query.

### Screenshots
image
image
image
image
image
image

### DIFF
--- a/babybuddy/templates/babybuddy/base.html
+++ b/babybuddy/templates/babybuddy/base.html
@@ -44,22 +44,6 @@
         <script src="{% static "babybuddy/js/vendor.js" %}"></script>
         <script src="{% static "babybuddy/js/app.js" %}"></script>
         {% if user.is_authenticated %}<script>BabyBuddy.PullToRefresh.init()</script>{% endif %}
-        {% block javascript %}
-            <script type="application/javascript">
-                function rememberAdvancedToggle(event, element){
-                    localStorage.setItem("advancedForm",event.newState)
-                }
-
-                window.addEventListener('load', function () {
-                    if(localStorage.getItem("advancedForm") !== "open"){
-                        return;
-                    }
-
-                    document.querySelectorAll("form details").forEach(function(node){
-                        node.open = true
-                    });
-                })
-            </script>
-        {% endblock %}
+        {% block javascript %}{% endblock %}
     </body>
 </html>
--- a/babybuddy/templates/babybuddy/base.html
+++ b/babybuddy/templates/babybuddy/base.html
@@ -44,6 +44,22 @@
         <script src="{% static "babybuddy/js/vendor.js" %}"></script>
         <script src="{% static "babybuddy/js/app.js" %}"></script>
         {% if user.is_authenticated %}<script>BabyBuddy.PullToRefresh.init()</script>{% endif %}
-        {% block javascript %}{% endblock %}
+        {% block javascript %}
+            <script type="application/javascript">
+                function rememberAdvancedToggle(event, element){
+                    localStorage.setItem("advancedForm",event.newState)
+                }
+
+                window.addEventListener('load', function () {
+                    if(localStorage.getItem("advancedForm") !== "open"){
+                        return;
+                    }
+
+                    document.querySelectorAll("form details").forEach(function(node){
+                        node.open = true
+                    });
+                })
+            </script>
+        {% endblock %}
     </body>
 </html>
--- a/core/forms.py
+++ b/core/forms.py
@@ -277,6 +277,14 @@ class NoteForm(CoreModelForm, TaggableModelForm):
 
 
 class SleepForm(CoreModelForm, TaggableModelForm):
+    fieldsets = [
+        {
+            "fields": ["child", "start", "end", "nap"],
+            "layout": "required",
+        },
+        {"layout": "advanced", "fields": ["notes", "tags"]},
+    ]
+
     class Meta:
         model = models.Sleep
         fields = ["child", "start", "end", "nap", "notes", "tags"]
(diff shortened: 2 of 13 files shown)


babybuddy/templates/babybuddy/base.html
```diff
@@ -47,6 +47,23 @@
...
+    window.addEventListener('load', function () {
+        if(localStorage.getItem("advancedForm") !== "open"){
+            return;
+        }
+
+        document.querySelectorAll("form details").forEach(function(node){
+            node.open = true
+        });
+    })
+</script>
```
cdubz, line 66: This does feel a bit awkward but I think it works well enough for a simple use case. Lets move this code in to the babybuddy,js file and namespace though.

core/templates/forms/layouts/advanced.html
```diff
@@ -0,0 +1,10 @@
+<div class="row">
+  <div class="row mb-3">
+    <details ontoggle="rememberAdvancedToggle(event)">
+      <summary>Advanced</summary>
+      {% for field in fieldset.fields %}
+        <div class="row">{% include "babybuddy/form_field.html" %}</div>
+      {% endfor %}
+    </details>
```
cdubz, line 8: My only objection here is that the details element feels very _not_ mobile friendly -- it's a small target link to click and it's right above the submit button. Any thoughts on how we would style to improve this?

core/forms.py
```diff
@@ -201,6 +249,7 @@ class Meta:
         widgets = {
             "child": ChildRadioSelect,
             "start": DateTimeInput(),
+            "type": PillRadioSelect(),
```
cdubz, line 252: Lets implement the same on the dropdown fields in `Feeding`?

core/forms.py
```diff
@@ -107,6 +107,31 @@ def save(self, commit=True):
...
+                "layout_attrs": fieldset.get("layout_attrs", {}),
+                "fields": [],
+            }
+            for field_name in fieldset["fields"]:
+                hyrdrated_fieldset["fields"].append(bound_field_dict[field_name])
+
+            hydrated_fieldsets.append(hyrdrated_fieldset)
+
+        return hydrated_fieldsets
+
```
cdubz, line 134: Hm yeah this seems... hacky. But I'm also unsure what other approach would work here. Regardless lets implement on the other models as well. I'm good with the idea generally.

cdubz (changes requested):

core/templates/forms/layouts/advanced.html
billybonks: Agree i can add some styles to improve it, will add a screenshot when done

billybonks: image
image
image
image

billybonks: https://github.com/babybuddy/babybuddy/pull/760/commits/b315b1e3dce11920cd29cfc80933c5de41eeb62c

core/forms.py
billybonks: Done

core/forms.py
billybonks: My issue with the current api is that It will be confusing for a new person 

In order to add a field, you must 

add it to fields in meta, then add it to field sets only then will it show

if something is added to fieldsets and not to fields then it will do nothing.

i think ideally the api should be

```
class SleepForm(CoreModelForm, TaggableModelForm):


    class Meta:
        model = models.Sleep
        fieldsets = [
            {
                "fields": ["child", "start", "end", "nap"],
                "layout": "required",
            },
            {"layout": "advanced", "fields": ["notes", "tags"]},
        ]
        widgets = {
            "child": ChildRadioSelect,
            "start": DateTimeInput(),
            "end": DateTimeInput(),
            "notes": forms.Textarea(attrs={"rows": 5}),
        }
```

billybonks: gave it a try but will require much deeper work to do it properly

image

billybonks:
@cdubz seems like an unrelated test failure https://github.com/babybuddy/babybuddy/actions/runs/7811519906/job/21306747076?pr=760#step:8:42

passed locally

cdubz:
> @cdubz seems like an unrelated test failure https://github.com/babybuddy/babybuddy/actions/runs/7811519906/job/21306747076?pr=760#step:8:42
> 
> passed locally

Yeah that is an unreliable test I need to sort out at some point. It fails randomly. Seems to be some issue related to timezones.

cdubz (changes requested):
I think I have just one last complaint here -- the radio buttons (see Type and Method in this example) need some more spacing. They seem to be right up against each other right now.

Screen Shot 2024-02-14 at 20 00 39

There may be a better way to lay them out generally but just added some evened out spacing would be a nice enough tweak here.

billybonks:
@cdubz, i tried to make them take up all available space, with spacing.

image

image


My only concern was that the buttons become quite huge on desktop.

image

so i only grow the buttons on mobile

image

cdubz:
Looks good! Thanks for you work on this!

cdubz (approved):

Issue #199, opened 2021-03-07
Unable to run on docker ARM64


markus-seidl:
I've tried to run the docker-compose.yml on an ARM64 device, but got 'Invalid exec format' upon startup.

The system is:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic
```

```
$ lscpu
Architecture:        aarch64
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  1
Core(s) per socket:  4
Socket(s):           1
Vendor ID:           ARM
Model:               4
Model name:          Cortex-A53
Stepping:            r0p4
CPU max MHz:         1296.0000
CPU min MHz:         408.0000
BogoMIPS:            48.00
Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32
```

After that I've tried building the image myself. This can be "easily" done with a modification of the docker-compose.yaml (leaving this here for others, if needed. Checkout the code into ./code ):

```
# ...
  app:
    image: babybuddy-arm64
    build:
      context: ./code/
# ...
```

Unfortunately, this yields the following error upon executing ```docker-compose build```

```
⠹ Locking..✘ Locking Failed!
ERROR:pip.subprocessor:Command errored out with exit status 1:
 command: /root/.local/share/virtualenvs/app-4PlAip0Q/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-resolver-axapuy0r/psycopg2/setup.py'"'"'; __file__='"'"'/tmp/pip-resolver-axapuy0r/psycopg2/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-resolver-axapuy0r/psycopg2/pip-egg-info
     cwd: /tmp/pip-resolver-axapuy0r/psycopg2/
Complete output (14 lines):
running egg_info
creating /tmp/pip-resolver-axapuy0r/psycopg2/pip-egg-info/psycopg2.egg-info
writing /tmp/pip-resolver-axapuy0r/psycopg2/pip-egg-info/psycopg2.egg-info/PKG-INFO
[...]

Issue #146, opened 2020-06-16
Deploy to Heroku Fails on Boto


gregcram-personal:
Here's the log from Heroku:

```
-----> Python app detected
 !     Python has released a security update! Please consider upgrading to python-3.7.7
       Learn More: https://devcenter.heroku.com/articles/python-runtimes
cp: cannot stat '/tmp/build_672264488fdeae79f3ff45554527477d/requirements.txt': No such file or directory
-----> Installing python-3.7.6
-----> Installing pip
-----> Installing dependencies with Pipenv 2018.5.18…
       Installing dependencies from Pipfile.lock (2c2ee3)…
       An error occurred while installing boto3==1.13.25! Will try again.
       Installing initially–failed dependencies…
       Collecting boto3==1.13.25 
         Using cached https://files.pythonhosted.org/packages/c7/13/abd70615a8b0a310ee105974eb35f5d94f87a66855e8d7447175dc6625e5/boto3-1.13.25-py2.py3-none-any.whl
       
       THESE PACKAGES DO NOT MATCH THE HASHES FROM Pipfile.lock!. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
           boto3==1.13.25 from https://files.pythonhosted.org/packages/c7/13/abd70615a8b0a310ee105974eb35f5d94f87a66855e8d7447175dc6625e5/boto3-1.13.25-py2.py3-none-any.whl#sha256=89ad0d7bb0a619fa5b10d4b23bba8f2f97ee500055038da2723a96d4f0f97447 (from -r /tmp/pipenv-nrx7r9c1-requirements/pipenv-qb9qovnj-requirement.txt (line 1)):
               Expected sha256 1c37c727e0e6f872f5262a893a1ed5422ef0f1ed6075d35b809e20c3072dad5b
                    Got        89ad0d7bb0a619fa5b10d4b23bba8f2f97ee500055038da2723a96d4f0f97447
       
       You are using pip version 9.0.2, however version 20.1.1 is available.
       You should consider upgrading via the 'pip install --upgrade pip' command.
       
 !     Push rejected, failed to compile Python app.
 !     Push failed
```

cdubz:
Boto likes to create as much complication for as a little value as it can manage, it seems. I'll take a look at this soon.

cdubz:
Looks like updating Python dependencies was enough here. Give it a try again and reopen if you still have an issue :+1:

Issue #41, opened 2018-05-04
sha256 of docopt does not match


novski:
when installing from `pipfile.lock` with `$ sudo pipenv install --three --dev` i get this error:

```
THESE PACKAGES DO NOT MATCH THE HASHES FROM Pipfile.lock!. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    docopt==0.6.2 from https://www.piwheels.org/simple/docopt/docopt-0.6.2-py2.py3-none-any.whl#sha256=0340515c74203895f92f87702896e45424bf51dc71bf15b4748450f50be04346 (from -r /tmp/pipenv-vf5_eub9-requirements/pipenv-k7_dvsro-requirement.txt (line 1)):
        Expected sha256 49b3a825280bd66b3aa83585ef59c4a8c82f2c8a522dbe754a8bc8d08c85c491
             Got        0340515c74203895f92f87702896e45424bf51dc71bf15b4748450f50be04346
```
can i do anything to workaround?

cdubz:
Hmmmm. I can't seem to reproduce this locally. Are you attempting this install on a pi (guessing based on #40)? What versions of `pip` and `pipenv` are installed?

I'm testing on my own Pi now and actually having trouble with this command as well, but not due to hashes. I think pandas does a source build and it seems to be taking a very long time on AR

cdubz:
This happens because the Pi uses wheel packages from [piwheels](https://www.piwheels.org/) so the checksums don't match. Using `--skip-lock` resolves this, but ARM has other issues. I'm going to close this issue and keep track in #43.

PR #761, opened 2024-02-06
lint: running black locally caused mater to fail

billybonks:
Seems to be passing in github, probably a version issue. Local version i am using is 24.1a1. @MrApplejuice was running into the same issue https://github.com/billybonks/babybuddy/pull/2

### DIFF
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -293,7 +293,9 @@ CSRF_TRUSTED_ORIGINS = list(
 # https://docs.djangoproject.com/en/5.0/topics/auth/passwords/
 AUTH_PASSWORD_VALIDATORS = [
     {
-        "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
+        "NAME": (
+            "django.contrib.auth.password_validation.UserAttributeSimilarityValidator"
+        ),
     },
     {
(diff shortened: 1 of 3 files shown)


billybonks:
seems like ci version 24.1.1 wants to change what my local version wants to do 24.1a1, tried to setup a lock file but failed @cben are you able to hlpe me solve this issue?

cdubz:
Hmmm I wonder why your local version wants 24.1a1? That's a fairly old alpha...

We don't commit a lock file due to differences between architectures. Things may have changed in the past few years to make this doable now but see #43, #41, #146, #199 for all the headaches that came with that.

billybonks:
I will try install a newer version. I got this via pipenv install.

cdubz:
I re-installed my env and did get the latest version. It doesn't make any formtting changes for the current `HEAD` ---

```shell
➜  babybuddy git:(master) pipenv run black --version
Loading .env environment variables...
black, 24.1.1 (compiled: yes)
Python (CPython) 3.10.12
➜  babybuddy git:(master) gulp format
[19:52:40] Using gulpfile ~/PycharmProjects/babybuddy/gulpfile.js
[19:52:40] Starting 'format'...
Loading .env environment variables...
All done! ✨ 🍰 ✨
176 files left unchanged.
[19:52:44] Finished 'format' after 3.47 s
```

billybonks:
I uninstalled then reinstalled works now thanks

Issue #759, opened 2024-02-06
Linting


billybonks:
What should I use to lint/formatting:

JS
JSON
Django templates

right now I am using prettier JS and JSON and DJLint for Django. but every file I touch gets reformatted.

cdubz:
The only linting that happens is via black and the `gulp lint` and `gulp format` commands. Anything else is not linted.

billybonks:
Any objecting to me adding the ones mentioned here? i am not sure whats the best option for Django templates

cdubz:
No objections. May be better to isolate that in to a separate PR though.

billybonks:
Agreed

Kind Regards
Sebastien Stettler

billybonks:
all prs are merged

PR #762, opened 2024-02-06
Add prettier

billybonks:
Follow up to this issue https://github.com/babybuddy/babybuddy/issues/759


.prettierignore
```diff
@@ -0,0 +1,8 @@
+vendor
+**/static
+*.html
+*.css
+*.md
+*.yaml
+*.yml
+*.scss
```
cdubz, line 8: Is there a reason we _shouldn't_ use prettier on these file types as well?

billybonks: i added markdown and yml

html is done by djlint
css/scss is done by stylelint.

Otherwise we will have to deal with formatter wars if we have multiple

PR #768, opened 2024-02-07
bugfix: dashboard shows ago when less then 1minute

billybonks:
This pull request broke it again https://github.com/babybuddy/babybuddy/pull/725


cdubz:
🤦

PR #769, opened 2024-02-12
feat: create mobile friendly lists

billybonks:
Tables on mobile do not work well because you have to scroll left and right, this will allow us to view all the required information at once.

Before making a reusable template to use for all lists i wanted to see if this is something that the project would like to have. It is possible to rejig the table into cards but its way more ergonomic to just render a separate list of cards. 

i was thinking of adding notes as a details at the bottom of the card that a user could open and close. The delete button would go into the edit page. i assume people mostly look at the data and edit versus deleting; thats why i am happy to move it a bit deeper

image


cdubz:
> Before making a reusable template to use for all lists i wanted to see if this is something that the project would like to have. It is possible to rejig the table into cards but its way more ergonomic to just render a separate list of cards.

Yeah definitely interested. One important use case to address though will be multiple children and how to make the card look to capture that. Ideally it is consistent across all models. Otherwise we'll just need to identify the important information from each model to display in the card.

> i was thinking of adding notes as a details at the bottom of the card that a user could open and close.

Makes sense I think. Notes are pretty common across models so this would also be an important one to make consistent.

> The delete button would go into the edit page. i assume people mostly look at the data and edit versus deleting; thats why i am happy to move it a bit deeper

Ageed.

billybonks:
image

Added the avatar helper that you built

MrApplejuice:
> multiple children and how to make the card look to capture that

Just as an idea here: For the android app, I made it so that one would swipe the whole screen left or right and make sure that the last active child remains the default selected one. That seems to cover most use-cases of babybuddy, I would imagine, since one typically cares for a single child, or twins, making it easy to switch between one ore two children with a single button press.

Since swiping is a bit awkward to implement, I suggest some "flipper buttons" on the top as a simple alternative. You could even make the list expandable in a combobox so that can one select a child from a list...

image

billybonks:
I think thats a great idea. i will probably  reuse Pills i developed for edit forms for simplicty.

I do think the app should have a single child vs multi child mode.

Right now its multi child focused, i think we can setup something for single child then remove a bunch of stuff from the ui, i will probably do that in a separate pr

billybonks:
@cdubz, decide to just follow a similar pattern for notes on the original table to save some time :), hope it is okay

image

billybonks:
Creted an issue for the child flipper as its a global change in filters.html, so best to handle it seperately

cdubz (changes requested):
The child image gets squished too much when the note is long. Can we improve this?

image

Otherwise this looks good!

billybonks:
Changes made to handle the image edge case.

1. Fixed the size to 35px
2. when notes are added we align the image to the top of the card when there are no notes it is center aligned

image
image


let me know if you find any other funny behaviours

billybonks:
@cdubz i have also refactored the code, in preparation to reuse across lists can you review the last commit please, and let me know if you want me to change anything with the architecture before i move forward

core/templates/core/diaperchange_list.html
```diff
@@ -17,6 +17,10 @@ <h1>
         {% endif %}
     </h1>
     {% include 'babybuddy/filter.html' %}
+    {% include "mobile/base/list.html" with list_item_template="mobile/diaper_list_item.html" object_list=object_list %}
+    <div class="mobile-list">
+        {% for change in object_list %}<div class="row"></div>{% endfor %}
+    </div>
     <div class="table-responsive">
```
cdubz, line 24: This can be removed now, right? I think this is causing an empty second list div:

image

Once that's resolved this is good to go!

cdubz (dismissed):

billybonks:
@cdubz  sorry for the delay.  I fixed the issue you mentioned https://github.com/babybuddy/babybuddy/pull/769#discussion_r1514489344. i started to explore adding more lists to find issues and i did find some please see commentary on last commit.

in addition to those commentaries i found issues with the pumping experience. wanted to get your thoughts before i move forward.

@MrApplejuice how did you display the lists on your mobile app

image
image
image

cdubz:
@billybonks your latest changes look good me! I don't see any direct questions in your commits -- was there are issue you need some feedback on?

cdubz:
@billybonks let me know if you want to continue work on this or for me to pick it up. I think it's far enough long as-is that I could finalize it if you're not able to get back to it for now.

MrApplejuice:
> how did you display the lists on your mobile app

For what it is worth, because I was directly asked and because better late than never:

image

This is how I did it in the Android app. It is pretty much the timeline and only the timeline with infinite scrolling if you are scrolling to the bottom. The idea being that you do not really care about the category of what had happened, but about the most recent events. A strong color scheme plus iconography makes sure that one can hopefully check at-a-glance what has happened.

However, this makes the lists not super comparable. I would probably leave out the duration and just use the start and end times, because, personally found those to be a useful to see at a glance than the duration which I pretty much only would use for statistics of aggregation or similar.

PR #774, opened 2024-02-25
Add feeding method history

abog854:
Hi there,

First, thanks a lot for that project, it's very cool !

My partner is breastfeeding and wanted to have stats of how long she fed over the past days, and also which breast she used.
So I added a card similar to the diaper change history, but for feeding methods.

Let me know how it looks.


billybonks:
Hi, are you able to add screenshots?

abog854:
Here you go.

This is a standard one for breast feeding:

image

But it works for any method. In case there are a lot for a day, it's only gonna display the top 3 methods in the "progress bar":

image

For us, it mostly serves two purposes:
 * check which breast was most used for feeding, see how balanced it was over the past days.
 * view feeding stats over the past few days: number and time spent feeding.

Maybe we could have this for the feeding type as well, and ideally allow the user to configure which one they want to display in the dashboard, but that's probably out of scope for that PR? I'd be open to giving it a try, if my baby gives me enough time :D

cdubz:
> Maybe we could have this for the feeding type as well, and ideally allow the user to configure which one they want to display in the dashboard, but that's probably out of scope for that PR? I'd be open to giving it a try, if my baby gives me enough time :D

#34 is the oldest open issue in this repo (: Definitely out of scope for this PR but I would be more than happy to have a PR adding _that_ feature, hah. I have got a little steam behind it on a couple of occasions but never quite got it over the finish line.

abog854:
I wanted to make something generic, but I do agree it probably makes more sense to have it specialized for breastfeeding (and if not breastfeeding the card could hide, anyway).

So I changed it, let me know what you think:

image

cdubz:
Looks good. Thanks for contributing!

PR #778, opened 2024-03-02
deploy: rebuild static files for mobile friendly forms

billybonks:
Tried to deploy https://github.com/babybuddy/babybuddy/pull/760, then realised that i still need to update the static files


cdubz:
Oops good catch!

PR #783, opened 2024-03-22
Add shortcuts to manifest

ahosgood:
https://developer.mozilla.org/en-US/docs/Web/Manifest/shortcuts

It would be great to have a shortcut for a generic "Start timer" but I couldn't add that because it's done through a form post with CSRF.


cdubz:
Neat! Having not used these before -- do you have a (visual?) example of how they work? Beyond the MDN spec. Seems cool.

ahosgood:
Sorry - I don't have an example on any current, public project 😅

There are some examples on [web.dev](https://web.dev/articles/app-shortcuts) but I appreciate that's very much a "look at what they _could_ look like". I had trouble getting a dev env set up for this so I'm not sure how easy it would be for me to test.

cdubz:
Ok no worries. I'll see if I can get it tested with my dev. Thanks for contributing!

cdubz:
Looks like this works according to the devtools at least. It'll be easier to test once it's the demo so I'll merge it now.

Just noting here some warnings from the devtools:

Screenshot from 2024-03-28 04-56-17

PR #800, opened 2024-04-15
Re-add diaper change color and amount fields

### LINKED ISSUE #799: Diaper change color removed with 2.3.0
With the UI updates in 2.3.0, it's no longer possible to define the color during a diaper change. Can this be brought back? It's still a column in the diaper change report so I was expecting this might've been unintentional to have removed?

Thanks!

cdubz:
Fixes #799

### DIFF
--- a/core/forms.py
+++ b/core/forms.py
@@ -121,7 +121,7 @@ class CoreModelForm(forms.ModelForm):
 
         for fieldset in self.fieldsets:
             hyrdrated_fieldset = {
-                "layout": fieldset["layout"],
+                "layout": fieldset.get("layout", "default"),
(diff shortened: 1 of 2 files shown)


th0mcat:
I think this broke the new Add Bottle Feeding option.  Besides the two Amount fields, whenever I try to add a feeding in the past, it gives me the “ Another entry intersects the specified time period. ” error.

image

image

cdubz:
@th0mcat hmmm do you mean that the issue started with v2.3.1? Or did you experience it in v2.3.0 as well (if you used it)? This change _shouldn't_ have impacted that check at all but it's possible something in the broader set of v2.3.0 did. Either way could you open a new issue for this?

th0mcat:
@cdubz can do!

PR #810, opened 2024-04-24
Change sleep pattern report to show awake times

simwir:
**Changes**
This Pull Request changes the sleep pattern report to also report awake time.
- Awake blocks are given a green color.
- Mouse over label of awake block shows the same information as the asleep blocks, but with the text "Awake".
- Days without any registrations are still blank.

**Screenshot**
image

**Reason**
When using the app I was lacking information about the length of an awake stint. The Sleep Pattern report almost has this information, but you need to estimate the size of a blank spot to see how the length.

**Open for comments**
 I did not find any opinion in the source code about report colors, so I used a standard color from plotly. Alternatives are welcome.


cdubz:
Nice addition -- thanks for contributing!!

PR #814, opened 2024-05-03
Fix layout of diaper change form buttons on mobile

josevh:
A small fix for a layout issue I noticed since updating.

Before:
babybuddy-before

After (Mobile):
babybuddy-after

After (Wide viewport):
babybuddy-after-wide


cdubz:
Thanks for contributing! Tests are failing in master and this looks clear enough 👍

PR #831, opened 2024-06-03
"Next Sleep" dashboard widget

fuhrysteve:
One of the paid baby apps has a feature that estimates the "Next sleep" time. I found the paid version to be too expensive, so I spent some time this weekend throwing this together for babybuddy.

Please let me know your thoughts and if there is anything I can do to improve it!

Screenshot from 2024-06-03 07-17-35
Screenshot from 2024-06-03 07-18-59


core/models.py
```diff
@@ -182,6 +184,96 @@ class Child(models.Model):
...
+                return cls.THREE_OR_FOUR
+            if age_in_days < 210:
+                return cls.TWO_OR_THREE
+            if age_in_days < 365:
+                return cls.TWO
+            if age_in_days < 545:
+                return cls.ONE_OR_TWO
+            if age_in_days < 1095:
+                return cls.ONE
+            return cls.ZERO
```
cdubz, line 236: What is the basis for this data? I recall there there being a pretty decent standard for this back when I had younger kids. It would be good to document here what our sources are.

core/models.py
```diff
@@ -213,17 +305,78 @@ def name(self, reverse=False):
...
+        1 Month  15.5 hours, 8.5, 7 (31-90 days)
+        3 Months: 15 hours, 9.5, 4.5 (91-180 days)
+        6 Months: 14 hours, 10, 4 (181-270 days)
+        9 Months: 14 hours, 11, 3 (271-365 days)
+        12 Months: 14 hours, 11, 3 (366-545 days)
+        18 Months: 13.5 hours, 11, 2.5 (546-730 days)
+        2 Years: 13 hours, 11, 2 (731-1095 days)
+
+        Returns tuple (nighttime: timedelta, naptime: timedelta)
+        """
```
cdubz, line 337: Ditto here -- lets document sources for this timing.

requirements.txt
cdubz, line 1: Can we avoid these extra dependencies updates on the branch (beyond adding the new enum field dep)?

babybuddy/migrations/0033_alter_settings_language_and_more.py
cdubz, line 1: Can you re-do the migrations to put the currently pending migrations (languages and timezones) in one migration and the changes relevant to this PR in a second migration?

Tangent... some day I need to figure out how to _not_ have these constant damn timezone migrations 😬

cdubz (changes requested):

cdubz:
I have pretty steadfastly tried to avoid _predictive_ features in Baby Buddy for a while but this seems like an interesting idea and the approach so far looks reasonable. Thanks for bringing it up.

In addition to the review CRs I have a few questions to consider:

1. The field is labeled as "Number of naps" but the card is labeled as "Next Sleep" -- the app distinguishes (to some degree) between naps and sleep so how can this be made more clear? Does the logic account for both things on only for naps?

2. What is the behavior like when the "next sleep" time arrives? E.g. what does the card says at the point in time that has been calculated for the next nap and beyond?

3. What is the behavior of the card when the naps is set to 0 (or auto determines 0)?

When you feel this is out of draft state I'll also do more functional review (and could answers those second two items myself).

babybuddy/migrations/0033_alter_settings_language_and_more.py
fuhrysteve: Hmm, I believe that is already the case? the changes relevant to this PR are in this file:

https://github.com/babybuddy/babybuddy/pull/831/files#diff-2b7f948e553ae46e64fcfb28cb9d8554fdafde8c9f6c071226e840a65b37b647

let me know if i misunderstood

fuhrysteve:
Thanks so much for taking a look at this! It'll probably take me a week or three to get this branch stable enough to consider merging, but I appreciate your willingness to consider this despite an understandable hesitation to shove in _predictive_ stuff too hastily.

>     1. The field is labeled as "Number of naps" but the card is labeled as "Next Sleep" -- the app distinguishes (to some degree) between naps and sleep so how can this be made more clear? Does the logic account for both things on only for naps?

I had in my head that I would put some indication as to whether the "Next Sleep" should be a nap vs bedtime (see the `next_sleep_is_nap` variable) so I can add that in for sure!

>     2. What is the behavior like when the "next sleep" time arrives? E.g. what does the card says at the point in time that has been calculated for the next nap and beyond?

Just now I pushed up some changes (which might not work 100% yet) that partially addresses this.

My general approach thus far has been to assume, when in doubt, that there is missing data (i.e. if the wake window is 2hrs and the last sleep ended 6hrs ago, the child probably slept but no one logged it). In those cases I simply return `None` so that the widget effectively does not render, since a bad estimate is probably worse than no estimate at all.

>     3. What is the behavior of the card when the naps is set to 0 (or auto determines 0)?

In this case the `wake_window` would effectively be `sleep_needed_nighttime - 24`.

There is some fuzzy logic (which I just pushed up) that attempts to make the next sleep earlier tonight if the previous night's sleep was significantly shorter than `sleep_needed_nighttime` (or `sleep_needed_nap`). It needs some work still, but I think it is getting close!

I am dealing with a weird timezone bug that I do not yet have sorted out. For some reason it seems to be an hour off, as though it were not recognizing DST, even though I have everything set to `America/New_York`. I'll keep working on that and see if I can figure out what is going on

core/models.py
fuhrysteve: Honestly a combination of some article that I googled, github copilot, and chatgpt might have even had a say in it. I can do some digging and come up with a more reliable source, or at least link to a reasonable article.

cdubz:
All sounds good!

> If their next nap should start 10mins ago, though, the widget should probably have some adjusted language. I'll do some testing and thinking about this

Yeah something like this would be nice. But it also makes it a little awkward when, for example, the predicted time hits, the child goes down for a nap, and the caregiver starts a timer. At that point the card wouldn’t be adding much value.

> I am dealing with a weird timezone bug that I do not yet have sorted out. For some reason it seems to be an hour off, as though it were not recognizing DST, even though I have everything set to America/New_York. I'll keep working on that and see if I can figure out what is going on

Ugh. Let me know if you need any help troubleshooting. I feel like there is some deep seated configuration issue with Baby Buddy’s TZ handling that I’ve just never been able to quite resolve.

core/models.py
cdubz: Haha yeah let’s try to find a good source to rely on. I’ll do some looking as well. I try to use clearly well researched data from sources like government bodies or NGOs when possible.

babybuddy/migrations/0033_alter_settings_language_and_more.py
cdubz: Oops yes I see that now!

core/models.py
cdubz: Ok so here is a WHO source _and_ the important thing I know I was forgetting... the _number_ of naps per day doesn't really matter. The important thing is just the total amount of sleep time: https://www.who.int/tools/your-life-your-health/life-phase/newborns-and-children-under-5-years/making-sure-newborns-and-children-under-5-years-sleep-safely (and more deeply: https://iris.who.int/bitstream/handle/10665/325147/WHO-NMH-PND-2019.4-eng.pdf).

So the WHO says:

- 0 - 3 months: 14 – 17 hours
- 4 - 11 months: 12 – 16 hours
- 1 - 2 years: 11 – 14 hours
- 3+ years: 10–13 hours

So that's where this predictive stuff gets a little tricky... I don't really want the app to prescribe something that isn't backed by strong research and guidance (e.g., number of naps per day as opposed to number of hours of sleep per day).

Can we think of a way to re-imagine this feature with a focus on recommended hours of sleep per day?

cdubz:
> I am dealing with a weird timezone bug that I do not yet have sorted out. For some reason it seems to be an hour off, as though it were not recognizing DST, even though I have everything set to America/New_York. I'll keep working on that and see if I can figure out what is going on

This might be resolved on the next release (v2.6.1), in case you come back to this PR.

dagr1709:
Hello, is the extension still being worked on? I would be very happy to have this feature.

PR #833, opened 2024-06-10
Add Feeding Patterns report

### LINKED ISSUE #567: Feeding pattern
Hi, loving using baby buddy for my week old newborn, wondering if there’s a similar report to the ‘Sleep Pattern’ report, but for feeding instead?

Thanks

seanwatson:
Closes #567 

Largely a copy of the sleeping pattern report. Preview:
Screenshot from 2024-06-10 10-45-15

I'm haven't worked with localization files before, so I'm not sure what's needed from me if those have to be updated. I'm also not sure about whether or not I should be running `gulp updatestatic`, but I can easily add that if needed.

If you think there's value in generalizing a "Pattern report" I could try refactoring a bit so sleep pattern and feeding pattern share more code.

### DIFF
--- a/reports/graphs/__init__.py
+++ b/reports/graphs/__init__.py
@@ -6,6 +6,7 @@ from .diaperchange_intervals import diaperchange_intervals  # NOQA
 from .feeding_amounts import feeding_amounts  # NOQA
 from .feeding_duration import feeding_duration  # NOQA
 from .feeding_intervals import feeding_intervals  # NOQA
+from .feeding_pattern import feeding_pattern  # NOQA
 from .head_circumference_change import head_circumference_change  # NOQA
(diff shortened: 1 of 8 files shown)


cdubz:
This looks good -- thanks for contributing!

> I'm haven't worked with localization files before, so I'm not sure what's needed from me if those have to be updated. I'm also not sure about whether or not I should be running gulp updatestatic, but I can easily add that if needed.

There is some localization stuff needed here but I'll do that in a follow up commit.

> If you think there's value in generalizing a "Pattern report" I could try refactoring a bit so sleep pattern and feeding pattern share more code.

It would be nice to share more code between the two but this works as-is now so I don't see any reason to hold it up for refactoring. That can be a follow up PR if you or anyone else thinks it worth the time 😄

PR #839, opened 2024-06-25
[LANGUAGE] Little Language Upgrade

HB9HIL:
added some more translations to german po file and compiled the mo


cdubz:
Thanks for contributing!

PR #843, opened 2024-07-01
Add ability to configure CORS headers

### LINKED ISSUE #842: Feature Request: CORS headers
I'm throwing together a very simple frontend for my BabyBuddy instance, optimized for how we're using it (basically stripping out options/inputs irrelevant to us). I'm trying to hit the API, but I'm getting a CORS error.

It seems like there's no way to configure `Access-Control-Allow-Origin` and other CORS headers.

It would be great to have a `ALLOWED_ORIGINS` env var that would specify allowed domains to be making CORS requests from.

jkjustjoshing:
Fixes #842 

I *think* I did the installation of `django-cors-headers` correctly, but I'm not too familiar with Python environment setup and was having a bit of trouble installing dependencies. Make sure my change to `requirements.txt` is correct and that I don't need to make any other changes elsewhere.

### Changes
* Install `django-cors-headers`
* Read comma separated list of origins from `CORS_ALLOWED_ORIGINS` env var, and pass as list to `django-cors-headers`
  * `.strip()` to remove whitespace from each item, allowing spaces in the comma separated list
  * Add `.strip()` to `ALLOWED_HOSTS` value for convenience
* Add/update documentation

Let me know if there are any tests I should add. Though if you feel like writing them yourself feel free - not sure how long it'll take me to figure out the Python test environment.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -24,6 +24,7 @@ uritemplate = "*"
 whitenoise = "*"
 django-taggit = "*"
 django-qr-code = "*"
+django-cors-headers = "*"
 django-dbsettings = "*"
 django-import-export = "*"
 
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -37,6 +37,7 @@ INSTALLED_APPS = [
     "api",
     "babybuddy.apps.BabyBuddyConfig",
     "core.apps.CoreConfig",
+    "corsheaders",
     "dashboard",
     "reports",
     "axes",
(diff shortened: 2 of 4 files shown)


cdubz:
Thanks for contributing! This looks good.

And thanks for adding the `strip` improvement to `ALLOWED_HOSTS` as well.

Testing this would probably be a PITA (as is everything having to do with CORS) so I don't think anything else is needed here.

If you end up sharing code or a blog post or anything for your frontend please come back and share a link. I always love to see the cool things people do on top of Baby Buddy 😄

cdubz (approved):

jkjustjoshing:
@cdubz just trying to simplify some common tasks my wife and I have been using. 

image

Also, my wife wanted to know if she was behind or ahead of her pumping from the day before (since the built-in chart doesn't indicate if the 3rd pump of the day was hours before/after today). So I built this visualization (green line is today, previous days are in blue, fading out the farther in the past they are)

image

Here's the code https://github.com/jkjustjoshing/maddie-buddy

cdubz:
@jkjustjoshing cool, thanks for sharing! I'm going to add this to BBotW as an example of interacting with the API from a Remix frontend.

And I like that graph... would like that like to get a version of it in the main app...

PR #850, opened 2024-07-20
Fix minor UI bug with feeding pattern report

seanwatson:
A browser on one of my devices outlines each entry of the report in white. That leaves an awkward area at the end of each day where there is no outline unless there happens to be a feeding across the midnight boundary. This adds a new entry from the end of the previous feeding to midnight in order to make the outline consistent.

Before: 
before

After:
after


cdubz:
Thanks for the fix!

PR #851, opened 2024-07-20
Updated Norwegian translation.

NorseByte:
Updated Norwegian translation at Poeditor, and imported it to Baby Buddy.


cdubz:
Thanks for contributing!

Issue #844, opened 2024-07-04
Unable to create tags in forms


drewsteinacher:
Hi, I've set up my Baby Buddy instance via the [Home Assistant Add-on](https://github.com/OttPeterR/addon-babybuddy), and it works great overall! Thanks for making this useful tool!

I did run into a strange issue that I could not fully troubleshoot myself after digging through the code a bit.

I cannot add a new tag via any activity form, and when I try, I get the following message:
![image](https://github.com/babybuddy/babybuddy/assets/26029945/87179ab8-1f46-4ba2-b7b8-467f5db02049)

Looking at the console, it appears that it is making a GET request to the root of my local Home Assistant URL, which of course gives a 404:
![image](https://github.com/babybuddy/babybuddy/assets/26029945/8f92ddb3-54cf-4606-9a7b-0d32a67abc87)
```
http://homeassistant.local:8123/api/tags/?name=NewTag
```

Side note: The GET request is expected and is done before the POST request here in `tags_editor.js`:
https://github.com/babybuddy/babybuddy/blob/master/babybuddy/static_src/js/tags_editor.js#L236

Comparing it with other forms that work (e.g. feedings), it sends successful requests to the Add-on's URL:
![image](https://github.com/babybuddy/babybuddy/assets/26029945/bf05c6f8-b544-4d7a-9494-2a578420c824)
```
http://homeassistant.local:8123/api/hassio_ingress/C-qJPCYtQopDWtao3qE40kIWuGnvVl3PfJMCTpvSHys/feedings/4/
```

I suspect this to be a configuration issue of some kind, so I apologize if this is actually a bug for the Add-on repo, but it was difficult for me to isolate where the actual problem was.

Lastly, I'll note that I am able to make tags via the Database Admin area, but that workaround takes a bit longer and may not be an option for non-admin users.

cdubz:
Thanks for reporting. I do think this is a bug (of sorts) that should be handled by the base application but I'm not quite sure yet how...

The issue here is that the building of the API URL is not aware of and does not account for Home Assistant ingress handling. For API routes this [happens in exactly two places](https://github.com/search?q=repo%3Ababybuddy%2Fbabybuddy+%22%7B%25+url+%27api%22&type=code) (for this tag editor and the "API Browser" link in the menu).

It seems we need a way to have some context when creating those URLs but this is where I'm a little unsure about the best approach.

[...]

PR #852, opened 2024-07-24
Re-implement homeassistant ingress-server middleware

MrApplejuice:
**Note: This relies on the removal of these lines from addon-babybuddy**:

https://github.com/OttPeterR/addon-babybuddy/blob/9ffa58851974ef192bdc94ef49153c6eab4899e1/babybuddy/root/etc/nginx/templates/ingress.gtpl#L15-L22

These lines were integrated into the homeassistant middleware for babybuddy in different ways:

1. Most of the "rewrites" are happening by using `set_script_prefix` now. This is a more comprehensive solution than the original monkey-patch for building absolute uris. Also works for the REST-Framework!
2. Redirects are manually rewritten to include the ingress-url. Because of (1) this can lead sometimes to double-prefixing, depending on what redirect method is used. I did not feel like hunting down all the variations of redirects that can happen and just check for double-insertions in the rewrite code.
3. `"{static_url}` and `"{media_url}`-string patterns are filtered and rewritten in returned html-responses so that they point to the correct /static and /media urls, respectively. I could not think of a more stable different solution for this, so this is just a re-implementation of the nginx-rewrite rules.

Caveats:
- Streaming responses are not getting rewritten. I do not think that this is relevant though for babybuddy, and if it ever is, I am generating a warning that this might cause an issue.
- [x] Bug: (Recognized this while writing) ... it is possible to change the backend dynamically in homeassistant - meaning switching over from "ingress" to "bare" mode more or less transparently. However, the current `set_prefix_path` method does a static reconfiguration of the running instance. So I need to implement a more dynamic "reset" feature for the prefix-path in case one does not use the ingress service for a particular query. Might or might not be an actual issue, but also makes testing simpler, so I will do this.

Fixes https://github.com/babybuddy/babybuddy/issues/844


MrApplejuice:
I tried to address the remaining issue today. I think that I mostly succeeded (though not yet committed), however, there is some silly `500 Internal Server` error popping up when one tries to reach a non-existing api URL, specifically one that misses a trailing `/`. This should not be happening according to my analysis because the [CommonMiddleware](https://docs.djangoproject.com/en/5.0/ref/middleware/#django.middleware.common.CommonMiddleware) should issue an redirect on a 404 case like this one to the trailing slash version. But instead of the redirect, I get a 500 error. In debug-mode it works.

I cannot tell why it works in DEBUG mode and does not work in non-DEBUG mode. It is kind of complicated to observe as well, because this seems to be an interaction between the homeassitant middleware and something else. So I am still in debugging mode...

cdubz:
@MrApplejuice that sounds suspiciously like an issue that arises like this from Django because certain routes are different in the routing configuration when debug is enabled. Generally this has to do with media issues though.

MrApplejuice:
Hi... sorry for my sudden absence. I got ill, and did spent my time in bed instead of fixing issues. I am back out of bed now, nut need to get going again. Going to be picking this up again now.

MrApplejuice:
Quick update from my side: Debugging this is really time consuming and after I was ill I did not find enough time to fix this last issue. I am on vacation now for one week and I will not bring my dev-battle station on that vacation to fix this. The issue is still on my todo list however and, I did not forget about it. It needs to wait another week however until I am back.

cdubz:
Just drafting this for now as I'm finally catching up on things and trying to stay organized 😁

MrApplejuice:
Hey there! Of course, Drafting this was fine! Especially, because I was gone for what feels like a year again. Attemtping to fix the last issues now...

babybuddy/middleware.py
```diff
@@ -158,6 +145,82 @@ def __call__(self, request: HttpRequest):
...
+                        .replace(
+                            f"'{media_trunc}",
+                            f"'{x_ingress_path}{media_trunc}",
+                        )
+                    )
+                    filtered_headers = {
+                        key: value
+                        for key, value in response.headers.items()
+                        if not key.lower().startswith("content-")
+                    }
```
MrApplejuice, line 217: This part is necessary to make media and static urls work. There are different solutions for this, in theory, but plugins do not like them. So instead, filtering the returned text for any url encountered and fixing up the path, seems to be the only "scalable" way... 

So this is pretty much a re-implementation of the pattern replace-rules from the home assistant integration's nginx-rules.

MrApplejuice:
Ok... here is the deal. I think this is "as good as it gets", "pretty ok" and most importanatly "better than the current version".


Changes in a row:
- The "url rewrite" now is driven by the means of directly controlling `set_script_prefix`, a very deep core-feature of django. I tested this a lot. Seems to work very well, and fixes the problems relating to the REST-API not being reachable from JavaScript due to incorrectly rendered URLs (see below).
- Pattern replace rules for MEDIA and STATIC resource urls. Pretty much the pattern-replace feature that was part of the nginx config files before, but this time _"scoped"_ so that the javascript-urls do not get butchered anymore. 
    - This was the problem why the tag editor did not work ion home assistant for so long: We would try to create the correct URL, but then nginx would butcher that correctly generated URL with its over-eager rewrite patterns.

Concentrating the rewrites in the homeassistant extension for babybuddy allows us to test this easily.

## About the "One more bug"

([see above](https://github.com/babybuddy/babybuddy/pull/852#issuecomment-2247850399))

This remains unadressed. Mostly, it does not matter though, and while I do not like it, this is the main reason why this was not merged yet. Now, after yet another week and thinking big-picture, we should probably not let this fix leave in pending state just because of the minor redirect breakage. It only happens in the REST-API and that one is mostly visited by applications anyways. The redirect that is broken is one that adds a trailing `/` to the URL, should someone visit `http://localhost:8888/api/temperature/79` instead of `http://localhost:8888/api/temperature/79/`. It does still work in normal babybuddy, and if in "homeassistant mode" one uses the correct url with trailing slash, it all works, too.

So I submit this to review now :D Finally, after 3 months.

cdubz:
@MrApplejuice thanks so much for digging in to this! Lets give it a try.

Issue #826, opened 2024-05-19
Missing timer coverage for POST /api/pumpings


skjdghsdjgsdj:
It looks like, while you can create a pumping entry through the web interface from a timer, the API coverage lacks the ability:

```
curl -v -H 'Authorization: Token ...' -H 'Content-Type: application/json'
    -X POST -d '{"child": 1, "amount": 10, "timer": 327}' 'https://.../api/pumping/'
```
results in:  
```
{"id":18,"child":1,"amount":10.0,"start":"2024-05-19T16:41:49.978134-00:00",
"end":"2024-05-19T16:41:49.978145-00:00","duration":"00:00:00.000011","notes":null,"tags":[]}%
```
Note the duration is effectively zero, not what timer `327` was tracking.

While I could use the API to just read the start of a timer and calculate the end, then POST accordingly, it would be far simpler if the schema accepted timer IDs like feedings do, and like the web UI implies is possible.

Schema attached that shows a missing timer ID field.

skjdghsdjgsdj:
<img width="1029" alt="Screenshot 2024-05-19 at 5 03 35 PM" src="https://github.com/babybuddy/babybuddy/assets/37231318/a3918e55-adb4-4cd5-be1e-de66b5414f63">
GitHub in Firefox didn't attach the image so here's trying from Safari.

cdubz:
Thanks for flagging. I _think_ we just need to change the `PumpingSerializer` @ https://github.com/babybuddy/babybuddy/blob/7937b23dc481a71ada34e1cea2547c9f00e34df2/api/serializers.py#L120-L123

to extend `CoreModelWithDurationSerializer` instead of `CoreModelSerializer` (as `FeedingSerializer` does).

PR #861, opened 2024-08-12
Allow pumping sessions to be timed via API

frenata:
Closes: #826

I tested this locally by hand. It *seems* like there isn't generally API level coverage over the relevant functionality, i.e. auto-filling in start/end/duration based on a linked timer, but I'm happy to expand test coverage or extend existing tests on request, I might just need a touch more guidance, as I'm still learning the codebase.


cdubz:
@frenata thanks for contributing! Sorry for the long review delay 😬 

I went ahead and added the test because I wasn't actually sure how this is tested either 🙃 Thankfully, the API test cases support extension of the timer tests well -- I just had to add one line to the pumping API test.

PR #867, opened 2024-08-23
feat(SMTP): allow to specify sender name

### LINKED ISSUE #866: how to configure EMAIL_FROM
Hello, I'm searching how to configure the email_from when the email is sent ? 

It seems that the documentation doesn't talk about this ? 

Actually, it seems to use webmaster@localhost, that seems to be the default option on django mail : https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DEFAULT_FROM_EMAIL

So, maybe adding something like can be enough ?: 
```python
     DEFAULT_FROM_EMAIL = os.environ.get("EMAIL_FROM") or EMAIL_HOST_USER or ""
```
here : 
https://github.com/babybuddy/babybuddy/blob/566b6cc52f9abdb99362e314f0258eec8ba1add6/babybuddy/settings/base.py#L279-L288

PS : my first reflex was to try to set `EMAIL_HOST_USER` to try to setup it, that's why I add it in the `or` (my smtp doesn't need an authentication)

thib3113:
Allow to configure `DEFAULT_FROM_EMAIL` from [django email](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DEFAULT_FROM_EMAIL) . 

By default, it will populate the from header with `webmaster@localhost` (which can be refused by email provider / or hard to handle on relay based on `From` header ) . 

Because my first reflex was to use the `EMAIL_HOST_USER` to configure it (providers can allow sending only mails with the name of your user), I set it as a default value . 

Fix #866

### DIFF
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -286,7 +286,7 @@ if os.environ.get("EMAIL_HOST"):
     EMAIL_USE_SSL = bool(strtobool(os.environ.get("EMAIL_USE_SSL") or "False"))
     EMAIL_SSL_KEYFILE = os.environ.get("EMAIL_SSL_KEYFILE") or None
     EMAIL_SSL_CERTFILE = os.environ.get("EMAIL_SSL_CERTFILE") or None
-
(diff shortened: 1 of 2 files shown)


cdubz:
Looks good -- thanks for contributing!

PR #871, opened 2024-08-29
Update child_detail.html

### LINKED ISSUE #726: Timeline previous and next buttons not always clickable
Something about the layout of the timeline page makes the previous and next buttons unclickable. It seems that the child details row extends the whole screen and is "over" the buttons so they can't be clicked. This is likely related somehow to the Bootstrap 5 upgrade.

oelegeirnaert:
On larger screens, the date navigation buttons on the timeline are not working as expected because they are on a lower layer.

By adding the z-index: 2, we're moving up the layer, and the buttons are clickable.

Fixes #726


cdubz (changes requested):
Thank you! I noticed this a while back bit forgot about it.

Just to be consistent using BS utilities, can you adjust this to use the [`z-2` class](https://getbootstrap.com/docs/5.3/utilities/z-index/) instead?

cdubz (approved):

PR #886, opened 2024-10-09
fix: Prevent 500 error when child object not passed

jcgoette:
It's been awhile, so please double check me, but this should prevent 500/child.RelatedObjectDoesNotExist error when calling `validate_unique_period` without a child object provided.

### DIFF
--- a/core/models.py
+++ b/core/models.py
@@ -364,7 +364,7 @@ class Feeding(models.Model):
     def clean(self):
         validate_time(self.start, "start")
         validate_duration(self)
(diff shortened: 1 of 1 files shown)


cdubz:
@jcgoette 👋 Sorry for the delay getting to look at this. Do you repro steps I can test with?

cdubz (changes requested):
Needs repo steps and/or test -- I am unclear what the issue is here.

jcgoette:
Nearly 2 years old at this point. Just close it out.

cdubz:
💀

Issue #813, opened 2024-04-30
Add support for webhooks/callbacks


skjdghsdjgsdj:
When we use Baby Buddy, we interact with it a couple ways: the web interface and the API through various integrations (Android app, custom devices, etc.). What would be nice is support for webhooks in the API.

Use case: we log a feeding one of three ways:
* Directly in the web interface
* Via the Android app
* Using a custom-built ESP32 device that uses the REST API

But, _regardless of how we log the feeding_, we'd like to get a webhook callback to an arbitrary HTTP(S) URL. That would let us set up a timer (outside of Baby Buddy) that notifies us after a certain time interval like two hours "hey, it's been a while, a feeding time is coming up shortly."

That's a specific use case around feeding, but a generic publish/subscribe mechanism would be nice for things like feedings, changes, etc. Another random example: a webhook around changes could let us separately keep track of likely number of diapers used, and then we get notified "hey, you've logged _n_ changes which probably means _n_ diapers and you might be running low."

I searched the docs and existing issues and didn't see this implemented or requested yet.

cdubz:
I’m surprised there isn’t a more mature Django package for this. I’m open to this but it will take some research and consideration.

skjdghsdjgsdj:
A potential idea, and shoot it down as you feel: MQTT support?

For example, for a simplistic initial use case, define an MQTT broker in a config file. Then if _any_ of the REST endpoints are hit, publish the endpoint and its payload to a user-defined (config file or something) MQTT broker. I have a Mosquitto instance available, and if Baby Buddy was just dumb--in a good way--and republishes the REST requests it gets to the MQTT broker, then another service can subscribe to the broker and take over from there. It would be loosely coupled and not require Baby Buddy to actually send its own outbound HTTP messages to arbitrary endpoints, and MQTT is by its nature a very light protocol.

Of course there could be information exposure, so after a basic implementation, maybe the MQTT relaying could be locked down to specific endpoints, users, etc., but it would be a nice starting point.

The de facto standard library for MQTT in Python seems to be https://pypi.org/project/paho-mqtt/ which I've used before and was pretty straightforward, at least as a subscriber

cdubz:
Webhooks I have at least a passing familiarity with but MQTT is entirely new to me. This looks and sounds interesting but I'd probably have to lean on someone else contributing actual code here. As with webhooks -- I'm open to it, just can't say it's something I'll have the time to jump on. Though MQTT has piqued my curiosity for sure 😄

[...]

PR #894, opened 2024-11-01
Initial POC of emitting MQTT events when saving models

skjdghsdjgsdj:
https://github.com/babybuddy/babybuddy/issues/813

### DIFF
--- a/babybuddy/settings/production.example.py
+++ b/babybuddy/settings/production.example.py
@@ -31,3 +31,16 @@ MEDIA_ROOT = os.path.join(BASE_DIR, "../data/media")
 
 # SESSION_COOKIE_SECURE = True
 # CSRF_COOKIE_SECURE = True
+
+# MQTT
+# Emit events to this MQTT configuration when adding items to Baby Buddy.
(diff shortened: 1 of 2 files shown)


skjdghsdjgsdj:
Merging 2.6.1

cdubz:
@skjdghsdjgsdj I think this looks reasonable enough and I was able to test it successfully. I pushed a few small adjustments.

My biggest curiosity is whether we can improve `_as_json_payload` to be more general... maybe take a look at https://docs.djangoproject.com/en/5.1/topics/serialization/?

skjdghsdjgsdj:
I tried to use the Django serializers but my five minutes' worth of Django experience proved unhelpful. I do agree that my impromptu "serialization" is bound to have edge cases that fail--hence all the `if`s already--so this might be a better option. I can try again if you'd like. We should however be careful that the payload looks pretty generic and not Django-specific, so any Baby Buddy refactorings over time are less likely to cause backwards-incompatible payload changes.

cdubz:
I’ll take a look at the serialization. I’ve never used it before so now I’m curious 😁

PR #897, opened 2024-11-05
Fix: #865 remove Timezone flag from documentation

### LINKED ISSUE #865: Update timezone docs
Follow-up to #864. If the TZ setting doesn't do anything and users should use the setting for that, then the docs should reflect that.  

(I'm happy to contribute a PR, just can't get to it right now)

laingcc:
Fixes #865


cdubz (changes requested):
Thanks for contributing!

There is one more place this shows up. Could you remove it from here in your branch as well? https://github.com/babybuddy/babybuddy/blob/60fda51cf46182ec01592ee30bc9e34aa3108291/docs/setup/deployment.md?plain=1#L117

laingcc:
Done!

PR #903, opened 2024-11-11
Update zh-Hans translation

huangyq23:
Fixed form label translation for tag fields


core/forms.py
```diff
@@ -135,6 +135,7 @@ def hydrated_fielsets(self):
 
 class TaggableModelForm(forms.ModelForm):
     tags = TagField(
+        label=_("Tags"),
```
huangyq23, line 138: When overriding fields in Django forms we need to explicitly set field attributes

huangyq23:
cc @hejhj

core/forms.py
cdubz: Thanks, good catch!

cdubz (approved):

PR #906, opened 2024-11-14
Serbian language support

laxsrbija:
Hi, I've populated the missing translation strings for Serbian over on POEditor and exported them.

This pull request should contain everything except the rebuilt `static` folder, as I kept getting exceptions when I tried to execute `gulp updatestatic`. 

Please also double-check if the language can actually be selected and used before including it in a future release. 
Thanks for this wonderful project!


cdubz:
Thanks for contributing! I'll merge this in and tack a few commits if needed.

PR #908, opened 2024-11-19
Fix AttributeError caused by Home Assistant middleware.

SeraphimRP:
This resolves an issue I noticed in /user/add-device where it would try to reference `request.is_homeassistant_ingress_request` but it doesn't exist when Home Assistant support is disabled.

This preserves the early return in #899 but brings back setting the value to False so we don't have to rewrite validation wherever this reference is made.

### DIFF
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -143,6 +143,7 @@ class HomeAssistant:
                 request.headers.get("X-Hass-Source") == "core.ingress"
             )
         else:
+            request.is_homeassistant_ingress_request = False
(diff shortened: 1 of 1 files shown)


cdubz:
Paging @MrApplejuice 🔔 

I swear something like this already came up and we tweaked it, but now I can't find it and I'm confused :hurtrealbad:

cdubz:
OH WAIT I see you already accounted for that, @SeraphimRP!

cdubz (approved):
Thanks for the catch and fix!

PR #919, opened 2024-12-08
Enable better numerical input on mobile devices

huangyq23:
### Before
before
### After
after


cdubz:
This is a nice enhancement — thanks!

Issue #929, opened 2024-12-27
feature request: add food type field when feeding type is solid


josephclaytonhansen:
I'm looking to migrate to Baby buddy from Huckleberry, BB looks amazing but it's missing one crucial feature. A solid feeding should have a field for what the solid food is. (It's important to keep track of potential allergen exposure count, etc.)

Huckleberry has a great UI for this, but even just a text field would be fine. What would be ideal would be a dropdown list/text field where you could either pick from previous solid foods or type a new one.

skjdghsdjgsdj:
It's not an ideal solution, but in the meantime you could use tags for this.

josephclaytonhansen:
That's not a bad idea, at least for me, but I think it'll be harder to sell my wife on that as opposed to the ease of HB ;) 

I may work on a PR myself, I feel like this would be fairly simple to set up

PR #930, opened 2024-12-27
add solid food type field to feeding

josephclaytonhansen:
See: https://github.com/babybuddy/babybuddy/issues/929
I was able to add a solid food form field, but I'm not sure how to get it to display only when the feeding type is solid_food.  I looked through the forms templates and the Python view files but didn't see an immediately obvious way to do this. Help is appreciated- this is 90% of the way there :)


core/forms.py
```diff
@@ -247,15 +247,56 @@ class Meta:
...
+        ("orange", "Orange"),
+        ("grape", "Grape"),
+        ("potato", "Potato"),
+        ("corn", "Corn"),
+        ("peanut_butter", "Peanut Butter"),
+        ("pumpkin", "Pumpkin"),
+        ("watermelon", "Watermelon"),
+        ("cantaloupe", "Cantaloupe"),
+    ]
+    solid_food_type = forms.CharField(required=False,widget=forms.TextInput(attrs={'list': 'solid_food_choices'}))
```
cdubz, line 288: I'd rather not maintain a list of prefill options here. Also, this won't work without an associated `datalist` element with those values.

core/models.py
```diff
@@ -322,6 +322,12 @@ class Feeding(models.Model):
         max_length=255,
         verbose_name=_("Method"),
     )
+    solid_food_choice = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True,
+        verbose_name=_("Solid Food Choice"),
+    )
```
cdubz, line 330:
```suggestion
    solid_food_type = models.CharField(
        max_length=255,
        blank=True,
        null=True,
        verbose_name=_("Solid food type"),
    )

```

core/migrations/0036_feeding_solid_food_choice.py
cdubz, line 1: This migration will need to be regenerated after applying the suggested field name change (to `solid_food_type`).

cdubz (changes requested):

Merinorus:
I'm dubious about the prefilled choices, but if the user can customize this and if it's shown on some reports, this would be a great feature to manage introduction of allergens!

PR #933, opened 2025-01-02
fix: weight and height percentile plotting error due to date range

jpzheng:
When a measuring date for weight/height is on or after the last available date in the WHO percentile data, plotting errors out. This fix limits the date range to either the last measuring date or the last WHO percentile data date, whichever is earlier.


cdubz:
Thanks for the catch and fix!

PR #934, opened 2025-01-03
Fix database url configuration option not functional

Thor77:
providing DATABASE_URL was ignored and providing typo'd DATABSE_URL resulted in
```
WARNING:root:No DATABASE_URL environment variable set, and so no databases setup
```
and an empty, defunctional django database configuration so no need for backward compatability


cdubz:
Thanks for the catch and fix!

PR #944, opened 2025-01-13
Use feeding end times for diff calculation and on dashboard

Thor77:
Especially for longer feedings it makes more sense to use the feeding end time for display on dashboard to see on a quick glance when the last feeding ended (and when the baby might be hungry again). I also adjusted this in the feedings list so the end time is used for difference calculation.


skjdghsdjgsdj:
While logically it makes sense, I thought the _de facto_ standard for feeding spans is to measure start-to-start. If this changes, I suggest it be user-controllable at least to avoid unexpected changes.

Thor77:
@skjdghsdjgsdj I'll add a user setting so it can be toggled with the current behavior as default. Would you be fine with that?

skjdghsdjgsdj:
Works for me. Obviously @cdubz is the product owner.

Thor77:
Just as an addition, an app that display the last feeding time since feeding end: https://apps.apple.com/de/app/still-app-mein-baby/id1439575933

skjdghsdjgsdj:
I tried to find a good source on why start-to-start is commonly used for feeding tracking rather than end-to-start. I couldn't find an actual source like a medical journal or article, but I could find various discussions.

The general consensus seems to be start-to-start is used for breastfeeding specifically because you know for sure when the start is, but not necessarily the end. For example, the baby could fall asleep, not really be uptaking any milk, etc. Therefore, measuring from the start of each feeding gives you more reliable data. So, measuring end-to-start at face value does seem more intuitive because it appears to answer the question "how long since the baby last ate", but it turns out using the data this way doesn't necessarily answer that question.

This of course doesn't apply to other feeding methods, particularly for solid food, but for consistency's sake I advocate start-to-start for all cases. Having said that, clearly there are some people like you and others who want it to perform differently, hence the user control through a preference or setting.

cdubz:
I'm OK with a setting for this. The start-to-start choice essentially boils down to #78, "I remember that!". I think ultimately the time is not as important as the amount? That may be why there is no clear research on the matter. Timing is just more convenient that before and after weighing for a simple understanding.

Thor77:
I updated the PR to add a user setting allowing to toggle this. This adds quite a lot of complexity, please let me know if there's a better / simpler way to implement this.

As for the general topic: I see the point in preferring start to start calculations, but for my wife and I it's still more useful to see on a glance when the last breastfeeding session ended as an estimate for the start of the next / whether her crying can be caused by hunger or something else. But maybe that'll change with time...

cdubz (commented):
@Thor77 thanks for your contribution here -- despite my slowness, I am looking at this now 😁

I wonder if this should live at the server configuration level instead of the user level? I'm thinking that it makes more sense there because in theory you probably want all users of a given instance using the same basis, right?

I also wonder if we can tack on a similar change for the same configuration for sleep entries (see #1033).

Thor77:
@cdubz thanks for taking a look!
As there's no UI for system settings it might be harder to change for some people, but I agree, it should probably be consistent for all users. So maybe environment variables `TIME_DIFF_FEEDING_END=True/False` and `TIME_DIFF_SLEEP_END=True/False`?
Adding this for sleep entries as well is a good idea.

cdubz:
@Thor77 it's a little confusing but we do actually have a site settings config here with a frontend -- see https://github.com/babybuddy/babybuddy/blob/master/babybuddy/site_settings.py

Thor77:
@cdubz ah, right, i'll update the PR accordingly

Thor77:
Updated to use a site setting

cdubz (approved):

Counfhou:
Is there a different PR that can be followed specifically for the sleeping setting?

Issue #789, opened 2024-04-01
REVERSE_PROXY_AUTH with /api access


zjean:
Hi, I am using the babybuddy web app with REVERSE_PROXY_AUTH setup (with Authelia).
That works great.
One thing however: I would like te use the api with token based authentication. I did so by whitelisting the /api from my Authalia config.
However, it looks like the Babybuddy is expecting the reverse proxy auth header for the /api endpoint.
Is it possible to exclude this header check for the api?

cdubz:
This makes sense -- #790 needed a small fix. Thanks for contributing!

zjean:
Hmm, my solution in this PR does not work.
First of all the check for "api/" is incorrect, it should be "/api/".
But, next to that, I always get a 403 error, even when I am sure the 'return None' is called in the middleware.

My knowledge of Python/Django is almost non-existent, so I have no clue how to debug this.
Any suggestions?

zjean:
@cdubz 
It looks like something weird goes on in the authentication backends for the remote user on the /api path.
The request is denied, I think because of the remote_user being empty.

When I change the RemoteUserBackend in settings/base.py to AllowAllUsersRemoteUserBackend the API is accessible, however I am not sure whether this has any implications..

cdubz:
Hm yeah I'm not sure why that would make a difference based on Django's documentation for that backend: https://docs.djangoproject.com/en/5.0/ref/contrib/auth/#django.contrib.auth.backends.AllowAllUsersRemoteUserBackend

🤔

Issue #907, opened 2024-11-14
Feature request: SSO support


skjdghsdjgsdj:
You probably saw this one coming. I'm not sure if it's the same as #517.

I've been amassing a set of self-hosted services for my network, among which is Baby Buddy. In doing so, that means accumulating logins. While a password manager helps keep track of things, it doesn't help with management: what if I want to add a new user and give them access to all the things? To that end, I set up [Authentik](https://goauthentik.io/) as an SSO. I've successfully integrated it with some services like [Immich](https://immich.app/) and integrating it with Baby Buddy would be nice.

Example use case: I want to invite a family member to access all my family-related stuff. If my self-hosted services include a photo gallery, a family calendar, a recipe manager, Baby Buddy, etc., how can I give them a single login for all the things and that keeps them logged in as they navigate around? SSO is obviously the textbook use case for that.

Letting the SSO handle authentication also gets us some features for "free" like TOTP codes, passkeys, password reset flows, en masse user deactivations, etc. None of those would need to be implemented in Baby Buddy proper if the SSO did them.

Some caveats/notes:

* I believe Authentik can solve as a generic authenticator for a reverse proxy, but that's a band-aid solution to actual user management. Users logging in via Authentik or any other compatible SSO should result in actual user entities in Baby Buddy instead of just merely granting access.
* API access must continue to function. Would that still work as an API key associated to a user or proper service accounts (or both)?
* Would this affect user roles? Would the SSO, Baby Buddy, or both define actual authorization in addition to authentication?
* Would we allow both SSO and local logins?
* An average user likely doesn't want SSO. A default Baby Buddy install should still use local authentication. An administrator can add SSO login, or replace the built-in authentication with SSO, if they want.

This is probably a roadmap thing instead of a simple feature to implement unless it's [easy with Django](https://pypi.org/project/django-sso/). I am also far from versed on all the details of how SSO actually gets implemented, but it's obviously much more complex than the built-in Django authentication/authorization.

As I obviously have a functional SSO set up, I'd be happy to set up a test instance of Baby Buddy with bleeding edge code or a separate branch to test prior to an actual release.

cdubz:
I think you should be able to use the existing functionality from #517 for this. I'm a little unclear on "proxy provider" note in their documentation, but this seems to imply it's possible to do: https://docs.goauthentik.io/docs/users-sources/user/user_ref#additionalheaders

[...]

PR #947, opened 2025-01-26
LDAP support

bishtawi:
Properly addresses https://github.com/babybuddy/babybuddy/issues/202 by introducing LDAP authentication. With LDAP, issues like https://github.com/babybuddy/babybuddy-for-android/issues/67 and https://github.com/babybuddy/babybuddy/issues/789 do not exist.

I tested this by forking the LinuxServer Docker file and built an image that pulled in my Baby Buddy git repo containing my LDAP changes. Additionally, I tested logging into LDAP-enabled Baby Buddy with the Android app and everything works as expected.

The dependency `python-ldap` requires the OpenLDAP client libs installed. More info (including installation instructions of dependencies) can be found at https://www.python-ldap.org/en/latest/installing.html#build-prerequisites

I also reverted https://github.com/babybuddy/babybuddy/pull/790 as according to https://github.com/babybuddy/babybuddy/issues/789#issuecomment-2067770083 that solution does not work. I didnt think it made sense to keep code that is broken. I can add back the code if you disagree.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -8,6 +8,11 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Install dependencies
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo DEBIAN_FRONTEND=noninteractive apt-get -yq install build-essential libldap2-dev ldap-utils libsasl2-dev
     - name: Install Pipenv
       shell: bash
       run: pipx install pipenv


skjdghsdjgsdj:
If my understatement of LDAP is correct, and maybe it isn’t, this means Baby Buddy (or the Android app) would gather the username and password and send it to the LDAP server for verification. I see a few issues with this, not with this implementation per se but LDAP in general:

* The user doesn’t necessarily know they should use their SSO credentials as the domain mismatches the SSO domain, most likely. The login form for sure would look different than their generic SSO login form.
* Users in the directory who use TOTP MFA must append their code to their password in some way.
* Passkeys aren’t supported, nor are presumably worse MFA mechanics like SMS or email.
* All of the above issues make the use of password managers more complex or effectively impossible.

For context in my thinking: I use 1Password as my credential manager and Authentik as my SSO. My Authentik account uses a passkey for MFA. My setup therefore isn’t LDAP-compatible, best I can tell. In my defense I don’t use the Android app but rather the web client in mobile and desktop browsers.

The only fix I see requires client changes: an SSO redirect option for the web client or rendering the login in some kind of webview for the Android client. Both options deliberately make Baby Buddy “dumb” about the authentication mechanics, but also make it very flexible and I’d suggest actually more secure as it doesn’t need to ever see the actual password uses for SSO. It also means password managers know the domain used for the login form and can suggest the right password automatically.

I may be completely off base on all these thoughts, but I wanted to provide a scenario where the proxy auth, while far from ideal, still works better for SSO logins than LDAP for me. Also Baby Buddy is @cdubz’s product and not mine so it’s hardly for me to say what ends up being released.

cdubz:
While I am the primary maintainer I try not to think of Baby Buddy as “mine” 😁 So happy to hear considerations about this PR. This isn’t something I have ever used and I’m kind of unclear on the implementation and implications.

@bishtawi do you have any thoughts on @skjdghsdjgsdj’s feedback here?

skjdghsdjgsdj:
To clarify my opinion: I'm not opposed to LDAP specifically as I'm sure some people could use it. Obviously you can @bishtawi given you took the time to make the integration! But I *am* suggesting that LDAP is not a viable *substitute* for the proxy auth as it exists today.

Rather, if we were to get rid of the proxy auth, I'd suggest [OpenID Connect (OIDC)](https://docs.goauthentik.io/docs/add-secure-apps/providers/oauth2/). It seems to be more popular lately in webapps vs. LDAP, presumably due to the flexibility it affords. My vague understanding of LDAP is it's a popular solution for users with Active Directory or similar solutions.

I detailed my experience with the proxy auth integration including its pros and cons in #907. And again, I could be totally wrong on how LDAP works.

PR #968, opened 2025-03-26
feat(feeding_pattern): Color coded feeding methods

philiptzou:
This PR applies different colors for different feeding methods, so it can deliver more information.

Before:
Screenshot 2025-03-26 at 1 01 27 PM

After:
Screenshot 2025-03-26 at 1 01 56 PM


cdubz (approved):
Thanks for contributing!

PR #971, opened 2025-04-05
Initial commit of darkmode icon

simon-bd:


simon-bd:
@cdubz anything that needs to be changed here for you to be able to work with it ✌🏼

bachjessen:
@cdubz if you could point me in the direction so I could make it easier to implement I would be more than happy to

Issue #943, opened 2025-01-12
Incompatible with Python 3.13: `TypeError: The type of __firstlineno__ (int) is not a valid Value.`


inducer:
The underlying issue exists in `django-dbsettings`.

This PR fixes this for me: https://github.com/zlorf/django-dbsettings/pull/69.

cdubz:
Thanks for reporting this. At some point I need to figure out depending patching in Python. Also got a PR waiting on a dep fix.

PR #973, opened 2025-04-21
Add django-allauth support

juantxorena:
This is related to bug #907. Using [django-allauth](https://docs.allauth.org/en/latest/), add the possibility to register and login via SSO to every supported provider. There are still some things to fix, however, other than the documentation.

Disclaimer: I'm neither a python nor frontend developer, so I'm not sure what I'm doing. Most of the code (which is only frontend code) is shamelessly copied from [Tandoor](https://github.com/TandoorRecipes/recipes), a recipe app.

If there's a provided enabled, the login page will look like this:
image
It should list every enabled provided. I only tested with keycloak because it's what I use. Clicking that button will redirect to another page, which is very ugly because I don't know how to do frontend, much less as python templates, so this would need work before merging, of course:

image

Pressing the "Continue" button will redirect to the actual keycloak (in this case) login site, and after successful login, the user will be simply logged in:

image

Other than the aforementioned ugly login page, there are some things that might need to be fixed:

In the `INSTALLED_APPS` section of the `base.py` settings file, at the end, I commited the app `"allauth.socialaccount.providers.openid_connect"`.  I wanted this to be a user setting in the production (or development) settings file, but as I said, I don't know python, and my attempts for django to use both sections in both files failed. Maybe somebody can help in this topic?

About the use, one should simply configure allauth in the settings file [according to the documentation](https://docs.allauth.org/en/latest/socialaccount/provider_configuration.html). As an example, for my case, this is what was working for my keycloak installation (provided that there's a client for babybuddy there, it's configured correctly, etc., but this is outside the scope):

```
SOCIALACCOUNT_PROVIDERS = {
    "openid_connect": {
        "APP": {
            "provider_id": "keycloak",
            "name": "keycloak",
            "client_id": "babybuddy",
            "secret": "*****",
            "settings": {
                "server_url": "REALM_URL/.well-known/openid-configuration"
            },
        }
    }
}
```

Summarizing, with minimal changes, we can add SSO support via OpenID connect, Oauth2, or any other popular provider to the app.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,6 +7,7 @@ name = "pypi"
 boto3 = "*"
 dj-database-url = "*"
 django = "~=5.1.1"
+django-allauth = {extras = ["socialaccount"], version = "*"}
 django-axes = "*"
 django-filter = "*"
 django-imagekit = "*"
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -56,6 +56,10 @@ INSTALLED_APPS = [
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "django.contrib.humanize",
+    "allauth",
+    "allauth.account",
+    "allauth.socialaccount",
+    "allauth.socialaccount.providers.openid_connect",
 ]
 
 # Middleware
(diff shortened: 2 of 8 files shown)


juantxorena:
Two additional topics:
- I noticed that I commited my development.py file, and even worse, I added my client secret, url, etc. Sorry for that, everything is internal and I can recreate the secret so it's not important, but it was stupid from my part.
- The devcontainer setting doesn't work. It installs a python 3.13 image, which apparently is not compatible with babybuddy (see #943). I had to change the image in the devcontainer file to `"image": "mcr.microsoft.com/devcontainers/python:3.12",`. I didn't forget to not commit this, however.

PR #986, opened 2025-07-10
German translations

myxor:
German translations updated

It looks like POEditor export messed up the .po file a bit but maybe this is also fine.


dagr1709:
Hi, what was changed? I redeploy my docker image. In the actually version are still some translations issue. 
Can we go and fix them together? Maybe if you are interestet i can go and make some screenshots and tell you the right german translation.
Thx for your work!

cdubz:
It looks like this is too outdated now to play nicely with with other DE changes. I will get POEditor cleaned up...

PR #1029, opened 2026-03-29
fix: preserve Set-Cookie when Home Assistant ingress rewrites HTML

zachfeldman:
The HomeAssistant middleware rebuilds text/html responses to prefix static and media URLs. It copied only response.headers into the new HttpResponse; Django stores csrftoken and sessionid on response.cookies, so browsers never received those cookies and CSRF/login failed behind ingress.

Copy response.cookies onto the rebuilt response (same pattern as add-on patch).

This is needed for a full fix for https://github.com/OttPeterR/addon-babybuddy/issues/81 (the Home Assistant Add On, AKA App now, that bundles Baby Buddy into the Home Assistant UI) which I am working on a larger fix for here:
https://github.com/OttPeterR/addon-babybuddy/pull/88

Made-with: Cursor

### DIFF
--- a/babybuddy/middleware.py
+++ b/babybuddy/middleware.py
@@ -1,3 +1,4 @@
+import copy
 import logging
 from os import getenv
 from time import time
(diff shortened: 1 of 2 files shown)


cdubz:
Sounds and looks reasonable enough.

@OttPeterR does this align with your understanding of what needs to happen addon side?

OttPeterR:
Yes, this lines up, thanks for taking a look

cdubz (approved):

zachfeldman:
Nice, thanks @cdubz and @OttPeterR !

PR #1041, opened 2026-04-15
Add "Recent Notes" dashboard card

### LINKED ISSUE #1038: Add "Recent Notes" dashboard card
Notes are currently only accessible through the notes list view. There's no way to see recent notes from the child dashboard, which means caregivers have to navigate away to check if any notes were left by another caregiver.

A "Recent Notes" card on the dashboard would show recent notes in a carousel with the note text visible below, making it easy to see what's been recorded without leaving the dashboard. This is especially useful for households with multiple caregivers.

cmenning:
Adds a "Recent Notes" dashboard card that displays notes in a Bootstrap carousel with synced note text, respecting the existing dashboard data age filter.

## Changes

- New `card_note_recent` template tag in `dashboard/templatetags/cards.py`
- New `cards/note_recent.html` template with carousel and synced note text display
- SCSS styles for note card layout in `dashboard/static_src/scss/cards.scss`
- Card added to the child dashboard layout (positioned near the top)
- Compiled static CSS updated via `gulp updatestatic`
- Tests for the new template tag (populated and empty states)

## Testing

- `gulp test` passes (299 + 1 tests, 0 failures)

Closes #1038


dashboard/templates/cards/note_recent.html
cdubz, line 1: Adjust this implementation to follow the style of https://github.com/babybuddy/babybuddy/blob/master/dashboard/templates/cards/statistics.html -- this approach skips the block implementation support getting all the card content inside the carousel. That way we can avoid the inline JS here.

cdubz (changes requested):
Just one CR to refactor the card implementation a bit. Otherwise this looks good.

cdubz:
Closing in favor of #1114

PR #1057, opened 2026-05-18
chore: regenerate stale static assets for medication feature

playcations:
The static files needed more regenerating.


cdubz:
@playcations I was assuming this was the issue -- is it not? Why did you close this one?

playcations:
The issue is that static files needed to be regenerated but I pushed the pr before testing and don't have time to test it at this moment so I just closed it until I can make sure it works.

PR #1058, opened 2026-05-22
Make DB options a dictionary type to be handed to the DB engine

### LINKED ISSUE #1027: DB_OPTIONS generates a TypeError
When following the documentation listed here: https://docs.baby-buddy.net/configuration/database/#db_options

I set the DB_OPTIONS for a sqlite backend with the following: `DB_OPTIONS="{'init_command': 'PRAGMA busy_timeout = 5000; PRAGMA journal_mode=wal; PRAGMA synchronous = NORMAL; PRAGMA mmap_size=134217728; PRAGMA journal_size_limit=67108864; PRAGMA cache_size=2000;'}"`

Setting the db_options triggers an exception: `TypeError: 'str' object is not a mapping`

I thought it might be the string quotes around the option, removing did not appear to help. 

I believe it has to do with this line: https://github.com/babybuddy/babybuddy/blob/master/babybuddy/settings/base.py#L130

```python
    if os.getenv("DB_OPTIONS"):
        config["OPTIONS"] = os.getenv("DB_OPTIONS")
    DATABASES = {"default": config}
``` 

I believe, the OPTIONS field has to be a dictionary type, I found that by parsing the string into a dictionary it worked.

```python

    if os.getenv("DB_OPTIONS"):
        import ast
        config["OPTIONS"] = ast.literal_eval(os.getenv("DB_OPTIONS"))
    DATABASES = {"default": config}
```

Bear in mind, I'm not great at Python so I'm not sure if this is actually best solution.

oddy48:
This fixes #1027 

I set the DB_OPTIONS for a sqlite backend with the following: DB_OPTIONS="{'init_command': 'PRAGMA busy_timeout = 5000; PRAGMA journal_mode=wal; PRAGMA synchronous = NORMAL; PRAGMA mmap_size=134217728; PRAGMA journal_size_limit=67108864; PRAGMA cache_size=2000;'}"

Setting the db_options triggers an exception: TypeError: 'str' object is not a mapping

Converting to the dictionary fixes this. I tested with postgres as well -- not sure if there is a test suite you'd prefer me to run to validate this.

### DIFF
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -127,7 +127,8 @@ else:
     if os.getenv("DB_PORT"):
         config["PORT"] = os.getenv("DB_PORT")
     if os.getenv("DB_OPTIONS"):
-        config["OPTIONS"] = os.getenv("DB_OPTIONS")
+        import ast
+        config["OPTIONS"] = ast.literal_eval(os.getenv("DB_OPTIONS"))
     DATABASES = {"default": config}
 
 
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -1,3 +1,4 @@
+import json
 import os
 import dj_database_url
 
@@ -127,9 +128,7 @@ else:
     if os.getenv("DB_PORT"):
         config["PORT"] = os.getenv("DB_PORT")
     if os.getenv("DB_OPTIONS"):
-        import ast
-        
-        config["OPTIONS"] = ast.literal_eval(os.getenv("DB_OPTIONS"))
+        config["OPTIONS"] = json.loads(os.getenv("DB_OPTIONS", "{}"))
     DATABASES = {"default": config}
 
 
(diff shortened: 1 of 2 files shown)


babybuddy/settings/base.py
cdubz, line 2:
```suggestion
import json
import os
```

babybuddy/settings/base.py
```diff
@@ -127,7 +127,9 @@ def strtobool(val):
     if os.getenv("DB_PORT"):
         config["PORT"] = os.getenv("DB_PORT")
     if os.getenv("DB_OPTIONS"):
-        config["OPTIONS"] = os.getenv("DB_OPTIONS")
+        import ast
+        
+        config["OPTIONS"] = ast.literal_eval(os.getenv("DB_OPTIONS"))
```
cdubz:
```suggestion
        config["OPTIONS"] = json.loads(os.getenv("DB_OPTIONS", "{}"))
```

cdubz (changes requested):
@oddy48 thanks for identifying this issue and contributing!

I don't think we want to `ast.literal_eval` here as it adds a lot of potential failure paths.

`json.loads` makes more sense here. Give this a test and see if it fits the need.

oddy48:
Works for me! I updated the file per your request. I also updated the docs to make them json compliant.

cdubz (approved):
Thanks for catching the docs change, too!

PR #1060, opened 2026-05-23
chore: regenerate stale static assets

playcations:
So the issue I was running into is that I n=only tested with debug being true and not production so I never saw the issue initially. After regenerating all static files and retesting, it seems to have fixed the issue. I ran through all files and testing, it seems to all run smoothly. 

The medication merge (ddfa07ed) updated the unhashed static/.../app.css
and babybuddy.woff2 but the .gz siblings, hash-named copies, and
staticfiles.json manifest were left pointing at pre-medication content.
Result: in dev (whitenoise prefers .gz) and prod (manifest rewrites to
hashed names), the medication icon's CSS rule and font glyph never
reach the browser.

This commit runs `gulp updatestatic` per the contributing docs, with
the production storage backend, producing a fully consistent tree.


playcations:
@cdubz here we go

cdubz (approved):

PR #1061, opened 2026-05-25
fix: XFF processing when using a reverse proxy take 2

fvlaicu:
While tesing I observed that the reverse proxy ip address was again logged.
`AXES_IPWARE_META_PRECEDENCE_ORDER` gets initialized somewhere, setting it as None enables inheriting the upstream

### DIFF
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -373,14 +373,6 @@ AXES_COOLOFF_TIME = 1
 
 AXES_FAILURE_LIMIT = 5
 
-AXES_LOCKOUT_TEMPLATE = "error/lockout.html"
-
-AXES_LOCKOUT_URL = "/login/lock"
-
-AXES_COOLOFF_TIME = 1
-
-AXES_FAILURE_LIMIT = 5
-
 # `None` makes Axes use the default precedence order set by django-ipware.
 # See: https://github.com/un33k/django-ipware#precedence-order
 AXES_IPWARE_META_PRECEDENCE_ORDER = None
--- a/babybuddy/settings/base.py
+++ b/babybuddy/settings/base.py
@@ -377,29 +377,7 @@ AXES_LOCKOUT_TEMPLATE = "error/lockout.html"
 
 AXES_LOCKOUT_URL = "/login/lock"
 
-AXES_IPWARE_META_PRECEDENCE_ORDER = (
-    "X_FORWARDED_FOR",
-    # Load balancers or proxies such as AWS ELB (default client is `left-most` [`<client>, <proxy1>, <proxy2>`])
-    "HTTP_X_FORWARDED_FOR",  # Similar to X_FORWARDED_TO
-    "HTTP_CLIENT_IP",  # Standard headers used by providers such as Amazon EC2, Heroku etc.
-    "HTTP_X_REAL_IP",  # Standard headers used by providers such as Amazon EC2, Heroku etc.
-    "HTTP_X_FORWARDED",  # Squid and others
-    "HTTP_X_CLUSTER_CLIENT_IP",  # Rackspace LB and Riverbed Stingray
-    "HTTP_FORWARDED_FOR",  # RFC 7239
-    "HTTP_FORWARDED",  # RFC 7239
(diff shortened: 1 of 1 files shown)


cdubz:
https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-reverse-proxies says:

> `AXES_IPWARE_META_PRECEDENCE_ORDER`: The names of `request.META` attributes as a tuple of strings to check to get the client IP address. Check the Django documentation for header naming conventions. Default: `IPWARE_META_PRECEDENCE_ORDER` setting if set, else `('REMOTE_ADDR', )`

I'm understanding that as saying `AXES_IPWARE_META_PRECEDENCE_ORDER` should already be set to the default value of `IPWARE_META_PRECEDENCE_ORDER` when it is not explicitly set, right?

fvlaicu:
I don't see it being set anywhere except the readme:
https://github.com/search?q=repo%3Aun33k%2Fdjango-ipware%20IPWARE_META_PRECEDENCE_ORDER%20&type=code

Want me to set it up as `IPWARE_META_PRECEDENCE_ORDER` instead of `AXES_IPWARE_META_PRECEDENCE_ORDER`?

Both work in my tests

cdubz:
The default value from the underlying python-ipware package: https://github.com/un33k/python-ipware/blob/cab810ced72457abcc77ffd234be9cb0a75f7ec3/python_ipware/python_ipware.py#L30-L52

fvlaicu:
> The default value from the underlying python-ipware package: https://github.com/un33k/python-ipware/blob/cab810ced72457abcc77ffd234be9cb0a75f7ec3/python_ipware/python_ipware.py#L30-L52

that gets overridden here: https://github.com/un33k/django-ipware/blob/32cbe48412aa8971f148dffba6d058a36d9c78b0/ipware/ip.py#L19

cdubz:
Ok yes, I see now. I assumed it was allowing the default value to come through but it apparently intentionally is _not_ doing that.

For a change here, I don't really love just copy-pasting the full default list. I would prefer to replicate whatever pre-ipware behavior we had or find some way to "inherit" the default settings from ipware ourselves. Thoughts?

babybuddy/settings/base.py
```diff
@@ -377,6 +377,8 @@ def strtobool(val):
 
 AXES_LOCKOUT_URL = "/login/lock"
 
+IPWARE_META_PRECEDENCE_ORDER = None
```
cdubz: Nice! I assume `AXES_IPWARE_META_PRECEDENCE_ORDER = None` doesn't have the same effect? If it _does_, I would prefer we use that. If it _does not_, please add some detailed documentation in line here about what this settings does and why we are doing it this way (per our PR discussion).

cdubz (changes requested):

babybuddy/settings/base.py
fvlaicu: `AXES_IPWARE_META_PRECEDENCE_ORDER = None` works too!

cdubz: Nice, ok. I like that better. Though we should also make a comment about the _why_ here because this feels sort of like an anti-pattern. I'll add that in the merge commit and do a `v2.9.2` for this.

babybuddy/settings/base.py
```diff
@@ -377,6 +377,8 @@ def strtobool(val):
 
 AXES_LOCKOUT_URL = "/login/lock"
 
+AXES_IPWARE_META_PRECEDENCE_ORDER = None
```
cdubz, line 378:
```suggestion
AXES_COOLOFF_TIME = 1

AXES_FAILURE_LIMIT = 5

# `None` makes Axes use the default precedence order set by django-ipware.
# See: https://github.com/un33k/django-ipware#precedence-order
AXES_IPWARE_META_PRECEDENCE_ORDER = None

AXES_LOCKOUT_TEMPLATE = "error/lockout.html"

AXES_LOCKOUT_URL = "/login/lock"
```

cdubz (changes requested):

cdubz:
Oops I messed that up lol. Will fix...

cdubz (approved):

PR #1065, opened 2026-05-28
Enhance Dockerfile and CI/CD for Baby Buddy with PostgreSQL support, ROOTLESS DOCKER IMAGE

ksullivan86:
this allows the creation of a rootless docker image so babybuddy can run completely rootless and adds a rootless Dockerfile. I have also included docker-compose files and a kustomization.yaml for k8s.


ksullivan86:
need to fix lint errors

PR #1066, opened 2026-05-28
Enhance Dockerfile and CI/CD for Baby Buddy with PostgreSQL support, ROOTLESS DOCKER IMAGE

ksullivan86:
this allows the creation of a rootless docker image so babybuddy can run completely rootless and adds a rootless Dockerfile. I have also included docker-compose files and a kustomization.yaml for k8s.


cdubz:
@ksullivan86 thanks for the PR -- can you speak to the _why_ of this change? I am not intimately familiar with Docker (and k8s not at all).

I have larger relied on the wonderful folks at LSIO for Docker support (see: https://docs.linuxserver.io/images/docker-babybuddy/) -- is this something that would live alongside that? Replace it? Enhance it? Etc.

PR #1076, opened 2026-06-25
fix: weight and height percentile sorting and slicing mismatch

### LINKED ISSUE #1072: WHO Percentiles are incorrect
Baby Buddy says <15 and the doctor is saying 25. I physically charted based off https://cdn.who.int/media/docs/default-source/child-growth/child-growth-standards/indicators/weight-for-age/cht-wfa-boys-p-0-2.pdf?sfvrsn=4db44d93_12 and get >15 as well. It seem like the calculations Baby Buddy is doing is messing up somewhere. 

Data:
* Boy
* 9 months
* 8.0229 kg

Intecpsp:
### Description

This PR fixes a bug where WHO growth percentile curves (weight and height) are plotted with incorrect numbers (e.g., showing a child in a lower/higher percentile than they actually are), even though the curves themselves render smoothly.

### Root Cause
1. **Array Length Mismatches**: While the `x` (dates) array was sliced to only plot up to the last measurement date (e.g. 274 elements for a 9-month-old child):
   `dates = dates[: dates.index(last_date_for_percentiles) + 1]`
   the corresponding `y` arrays (e.g. `p15_weight`) were NOT sliced and retained their full length (1,858 elements). Passing mismatched `x` and `y` array lengths to Plotly causes it to incorrectly map the index coordinates, resulting in the wrong percentile numbers being aligned with each date.
2. **Unordered Query Results**: The database queries for `age_in_days` and the respective percentile values were executed without an explicit `.order_by()`. In databases like PostgreSQL, the absence of an explicit `ORDER BY` clause can cause rows to be returned out of order, further scrambling the alignment of dates and percentile values.

### Fix
* Added an explicit `.order_by("age_in_days")` to the percentile querysets in `reports/graphs/weight_change.py` and `reports/graphs/height_change.py` to guarantee chronological order.
* Sliced each `y` data array to the same `end_index` as the `x` (dates) array, ensuring consistent array lengths for Plotly.

Fixes #1072


cdubz (approved):
Thanks, @Intecpsp -- great catch!

PR #1078, opened 2026-06-26
Plot growth percentiles by corrected age for preterm children

pietrouk:
# Plot growth percentiles by corrected age for preterm children

## Summary

Adds optional corrected-age support to the weight and height growth charts.

When a child has a **due date** recorded that is later than their birth date
(i.e. they were born preterm), the WHO percentile curves are anchored to the
due date instead of the birth date, so they line up against the child's
**corrected age**. This is the standard way preterm growth is read for roughly
the first two years, and without it a preterm infant can look like they are
tracking below their percentile.

Closes/relates to the discussion where adding this was welcomed. Happy to
adjust the approach based on your preferences (a couple of open questions at the
bottom).

## Approach

The growth charts plot actual measurements by their real date, and position the
percentile curves by mapping each curve point's `age_in_days` onto
`birth_date + age_in_days`. The minimal, correct change for corrected age is to
anchor those curves to the **due date** instead:

- Actual measurements stay on their true calendar dates.
- The x-axis stays "Date".
- Only the reference curves shift, by exactly the gestational correction.

This is equivalent to the UK-WHO "draw the line back by the weeks of
prematurity" method, and it keeps the change small and the semantics of every
other chart untouched.

Correction is applied automatically whenever `due_date > birth_date`. A due date
on or before the birth date (term or post-term) applies no correction. When
correcting, the chart title gains a small subtitle: *"Percentiles plotted by
corrected age"*.

Head circumference and BMI charts are intentionally left alone, since they have
no percentile reference curves to realign.

### Also included: a small robustness fix

The percentile x-range was truncated with
`dates[:dates.index(last_date_for_percentiles) + 1]`. With curves anchored to
the due date, every measurement can predate the first (corrected) percentile
point, in which case that `list.index()` lookup would raise. It now slices by
comparison instead, which is safe in both the corrected and uncorrected cases.

## Changes

- **core**: optional `Child.due_date` field (+ migration), form field with a
  date picker, and the field exposed on `ChildSerializer`.
- **reports**: `weight_change` and `height_change` take an optional `due_date`,
  anchor percentiles to corrected age, and add the subtitle; both report views
  pass `child.due_date`.
- **tests**: model tests for the new field; weight-graph tests covering the
  corrected-age note, the no-correction (term / post-term) cases, and the
  early-measurement regression; updated the Child API expected payload.

## Testing

- `python manage.py test core reports api` → all green.
- `black` clean.
- `makemigrations --check` confirms the field is fully captured by the migration.

> Note on the migration: on a clean checkout, `makemigrations` currently also
> wants to emit unrelated `AlterModelOptions` for several models under the
> pinned Django version. That churn is pre-existing and unrelated to this
> change, so the migration here is deliberately limited to the single
> `AddField`.

## Open questions

1. **Automatic vs. toggle.** Correction currently kicks in automatically when a
   due date is set. Would you prefer a per-view toggle (actual vs. corrected)
   instead, or in addition?
2. **Due date vs. gestational age.** I went with a due date because it maps
   directly to corrected age and is a single field. An alternative is to capture
   gestational age at birth and derive the due date. Happy to switch.
3. **Subtitle gating / i18n.** The new strings are wrapped in `gettext`; let me
   know if you'd like me to run `makemessages` to update the catalogs, or if you
   handle translations separately.


pietrouk:
Hi @cdubz 👋

Just a friendly ping on this one, in case it slipped through the cracks. No pressure, I know there's been plenty going on (congrats on 2.10!).

I've been running this branch on my own instance since June and it's worked well, with no issues so far. The only downside is that I'm stuck on v2.9.2 to keep it, so I'd love to get it merged and catch up with the latest releases :)

It has picked up merge conflicts since I opened it. For one, its `0037` migration now clashes with `0037_alter_diaperchange_color`. I'm happy to rebase onto master and renumber the migration whenever you're ready to look at it. I'm also glad to change the approach if you'd prefer something else, like a toggle, or gestational age instead of a due date (see the open questions above).

Thanks for all your work on Baby Buddy!

cdubz:
@pietrouk go ahead and get it up to date -- I've been trying to catch up recently so strike while the iron is hot!

pietrouk:
Thanks @cdubz! I've rebased onto master and renumbered the migration to `0038_child_due_date`.

The only conflicts were in the weight and height graphs, where #1076 had fixed the same lines. I kept that fix and combined it with this PR's comparison-based cut-off, which is still needed because shifting the curves to the due date can put the last measurement before the first percentile point. All tests pass and black is clean.

pietrouk:
@cdubz rebased onto current master, so this is up to date again and CI is green.

No rush at all, but it would be good to get a look while it's fresh. That's two rebases now, and I'd rather not need a third 🙂

cdubz (changes requested):
Thanks, @pietrouk -- looks good and just a few CRs to tackle from the Copilot review.

pietrouk:
Thanks @cdubz! All five points from the Copilot review are addressed in 537165dd:

- `due_date` is now in the `ChildAdmin` field whitelist, so it can be managed from the admin like the other child details.
- The corrected-age subtitle is gated on percentile data, so the plain weight and height reports (which pass an empty percentile queryset) are no longer labelled as corrected.
- The weight tests now assert the first x-value of a percentile curve: the due date when corrected, the birth date otherwise. The previous tests would have passed even if the curves had stayed anchored to the birth date.
- `reports/tests/graphs/tests_height_change.py` mirrors the weight coverage for height, including the early-measurement edge case.

Full suite is green (439 tests) and black is clean.

pietrouk:
Small follow-up: CI caught that the migration number had collided with the new `0038_tummytime_notes` on master, so I rebased again and renumbered mine to `0039_child_due_date`. The renumbering is folded into the original commit, so neither commit leaves the migration graph with two leaves.

All checks are green now. That's the third rebase after all 🙂 — and since the collision recurs whenever a new `core` migration lands, it's ready whenever you have a moment.

cdubz (approved):

cdubz:
All good -- thanks for moving this one along, @pietrouk!

PR #1079, opened 2026-06-26
Add UK-WHO preterm weight percentiles for boys and girls

pietrouk:
# Add UK-WHO preterm weight percentiles for boys and girls

## Summary

Adds weight-for-age percentile reference data for the **pre-term** period
(roughly 23 to 40 weeks gestation), for both boys and girls. This fills in the
growth-chart reference curves for the weeks before term, which Baby Buddy's
existing WHO data does not cover (the WHO standard starts at term / age 0).

Builds on #PR1_NUMBER (corrected-age growth charts). That PR anchors the
percentile curves to the due date (corrected age 0 = the due date). This PR
supplies the negative-age rows that populate the region before the due date, so
a preterm child's chart shows reference bands across the weeks before term
rather than empty space.

## Why it depends on the corrected-age PR

The new rows use negative `age_in_days` (−119 to −1). They only make sense once
the percentile curves are anchored to the due date, which #PR1_NUMBER
introduces. Without that change the curves anchor to the birth date, so
negative-age rows would plot *before birth* and be meaningless. So this should
merge after #PR1_NUMBER, and the migration depends on `0037_child_due_date`.

## Data and verification

- Source: RCPCH `rcpchgrowth` UK-WHO neonatal reference data
  (`uk_who_neonate_male_weight_cubic_daily_lms.json` and the female
  equivalent), which is the UK1990-based preterm section behind the UK-WHO
  charts.
- Percentiles derived from the published L/M/S values with the LMS method at
  z = ±1.880794 (P3/P97), ±1.036433 (P15/P85) and 0 (P50), matching the
  z-scores behind the existing WHO columns.
- Every value was checked back against the `rcpchgrowth` source (max error
  below 0.01%).
- 119 daily rows per sex (238 total), days −119 to −1. The migration is
  reversible.

I used UK-WHO rather than WHO because WHO has no sub-40-week standard. If a
different reference is preferred (see open questions), the same method can
regenerate the data.

## Testing

- Migration applies cleanly on top of `0037_child_due_date`.
- Both sexes load with the expected values, and a preterm child's weight chart
  renders against the new curves with corrected-age anchoring.
- `black` clean; `makemigrations --check` reports no model changes from this
  data migration.

## Open questions

1. **Licensing — please confirm before merge.** The RCPCH / UK-WHO reference
   data has its own terms, and Baby Buddy is BSD-licensed. Please confirm that
   redistributing these derived values in the project is permitted. If it
   isn't, that decides the source for us and I'll switch (see below).
2. **Reference choice.** UK-WHO suits UK users and matches the corrected-age
   PR's intent. If you'd rather have an international prescriptive standard in
   the WHO spirit, INTERGROWTH-21st is the candidate; Fenton 2013 is the other
   widely used preterm reference. Happy to regenerate from either.
3. **Term boundary.** UK1990 term medians are slightly higher than WHO's, so
   joining this data (ending at day −1) to the existing WHO data (starting at
   day 0) leaves a small step in the bands at the due-date line. The official
   UK-WHO charts avoid this by keeping the neonatal reference until +2 weeks and
   only then switching to WHO. Doing that here means also adjusting the WHO rows
   for days 0–13, which I've left out to keep this PR to a single concern. Let
   me know if you'd like it folded in.
4. **Scope.** Weight only. The same source has neonatal height and
   head-circumference references; I can add those as a follow-up.


pietrouk:
this PR is dependent on [1078](https://github.com/babybuddy/babybuddy/pull/1078)

PR #1080, opened 2026-07-07
Fix timeline issues across days

### LINKED ISSUE #1059: Inconsistent display behavior for timeline events that cross days
Another minor issue in the medication section that I noticed. 

It's now 01:00 23-05-2026 local time (Europe/Amsterdam - baby buddy has the same TZ), and I saw the attached in the timeline. 

The moment the medication was taken is correct, but the (fundamentally very useful) notification of when the next dose is needed 6h later is shown on the 22nd, while actually it needs to be on the 23rd (which is just empty right now). 

Perhaps this is something that will resolve itself when it actually becomes 04:31 local time, but it's a bit weird now. 

image

oddy48:
This fix partially resolves #1059 by at least putting start and end dates on the correct day. For example, if sleep starts on 1/1 and ends on 1/2, you will see the falling asleep on 1/1 and the wake up on 1/2.

This code introduces an issue with one of the tests, I'll mark this PR as a draft until I can figure out how to resolve.

Previous:

Screenshot_20260707_163819

With this change:

Screenshot_20260707_163628

Screenshot_20260707_163245


oddy48:
Everything looks good. The failing test is for a file I didn't change. Not sure your preference for how to handle. 

Also, let me know if you prefer I squash my commits or you'll do on merge.

oddy48:
@cdubz thoughts?

cdubz (dismissed):
@oddy48 thanks for working on this! Two things --

1. Can you compare you work to #1088? Just to see if there are any gaps between the two (this PR looks cleaner, if nothing else).
2. Can you add a test that covers this change?

oddy48:
1. It appears #1088 is almost exactly the same as this fix (albeit with some other files linted). I'm not sure why the code-diff shows it as a whole-scale replacement, but to my eye it appears that they did almost the exact same fix but created helper functions for those. If I had to nit-pick, it appears that their fix does not account for the next medication time. Functionally, this PR and that are equivalent.
2. Happy to add a test to cover this change -- I'm not super familiar with Python in general, but I can take a swing at it. A couple questions:
- The test data only has entries that go over midnight for sleep, all the other duration based fields are single day (tummy time, feeding), and there doesn't appear to be test data for medicine. Would you like me to add data for all the test scenarios or should I only test the situations which are covered in the existing test data (e.g. only sleep)?
- I see you have the tests here: https://github.com/babybuddy/babybuddy/blob/master/core/tests/tests_views.py#L213 is this where you'd want me to add additional tests? 
- I will likely rely on AI assistance for helping me write the tests, Python is not my primary language. Do you have any concerns with that?

cdubz:
> I will likely rely on AI assistance for helping me write the tests, Python is not my primary language. Do you have any concerns with that?

No concerns. I'm not even totally sure what a test _should_ look like here. So opportune for AI assist 👍

oddy48:
Alright - so AI suggested I don't test against the view directly - instead I should add the objects I need and call the get_objects function directly, the tests look good, they test cross boundary checks well. 

Somewhat relatedly,  I had originally wanted to test the following cases:

1. One day (< 24 hours)
2. across midnight
3. midnight test (start or end at midnight)
4. >24 hour test (e.g. 2 or more days)

The LLM assured me that that was overkill - so I only test the second case which is the critical one.

Thoughts on the design and the tests?

cdubz (approved):
All good -- thanks for the fix, @oddy48!

PR #1082, opened 2026-07-08
Add support for bidirectional language in base.html

MosheRW:
This pull request updates the base HTML template to better support internationalization and right-to-left (RTL) languages. The most important changes are:

Internationalization and RTL support:

* [`babybuddy/templates/babybuddy/base.html`](diffhunk://#diff-7e06abb07b32de9bc8d9fa52ee76bb73e77f5cb299d173978dc874cadb4e35b9R3-R7): The `lang` attribute for the `<html>` tag is now dynamically set based on the current language, and the `dir` attribute is set to either "rtl" or "ltr" depending on whether the current language is right-to-left. This improves accessibility and layout for users of RTL languages.

before:
image

after:
image


cdubz (approved):
Thanks, @MosheRW!

PR #1085, opened 2026-07-10
Fix previous sleep calculation (#1006)

### LINKED ISSUE #1006: Use last end time when calculating "Previous" on sleep page
The sleep page table includes a "previous" column. The value of this column seems to be [n].start - [n-1].start

A value of [n].start - [n-1].end, when [n-1].end is available, would give the wake window. This value in the table would help me understand the length of wake windows, which I can't easily read from the sleep patterns report.

Of note, the last sleep widget setup (duration since last time entry) is very nice for our home assistant setup where a single press creates a sleep entry with zero duration, and a double press updates it with an end time. When kid is sleeping, we see how long they've been sleeping. When kid is awake, we see how long they've been awake.

Great software, cheers and thanks for making this!

gustavodiasgit:
## Summary

This PR fixes the calculation of the **Previous** value shown on the Sleep page.

Previously, the displayed time was calculated from the start of the previous sleep to the start of the current sleep.

This change calculates the interval from the end of the previous sleep to the start of the current sleep, matching the expected awake time between sleep sessions.

Fixes #1006

### DIFF
--- a/babybuddy/templates/babybuddy/nav-dropdown.html
+++ b/babybuddy/templates/babybuddy/nav-dropdown.html
@@ -15,356 +15,356 @@
             </a>
             <div class="d-lg-none d-md-none d-flex me-auto p-0 ms-2">
                 <div>
-                <a class="text-body-secondary"
-                   href="{% url 'dashboard:dashboard' %}"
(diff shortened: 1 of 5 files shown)


cdubz (changes requested):
@gustavodiasgit please update this one with the latest main branch changes and resolve conflicts.

PR #1088, opened 2026-07-14
Fix timeline cross midnight (#942)

### LINKED ISSUE #942: Timeline bug: activities finished after midnight will stick to previous day
Hey there!

This is probably one of the very firsts bugs I've seen in Baby Buddy, as it is so easy to see (and reproduce) that.

Let's take this first example: my baby sometimes sleeps earlier in the night, so she can wake up by... 00h20min. Let's say it happened today, so instead of showing she woke up today (January 10), her sleep will be shown instead and only in the timeline for yesterday (January 9).

In the first example, it's certainly something one could see just as a minor bug (20min, come on!). But the bug is "from 00:00:00 on", so if things play out better than expected(?) during a night you may have several hours shown as in the previous day's timeline:

image

I imagine this happens because the code assumes it should be like that whenever an activity start time is before midnight, with no filtering to split it for the timeline purpouses.

Best regards y saludos

gustavodiasgit:
## Summary

Fixes timeline display for sleep, feeding, and tummy time entries that span midnight.

Previously, both the start and end events were shown on the day the activity started. This change:

- selects duration-based entries when either the start or end falls within the requested day;
- displays each timeline event only on the day matching its timestamp;
- preserves feeding time-since-previous behavior.

## Testing

- Ran `black` on `core/timeline.py`
- Ran `python manage.py check`
- Manually verified entries spanning midnight

Fixes #942

### DIFF
--- a/babybuddy/templates/babybuddy/nav-dropdown.html
+++ b/babybuddy/templates/babybuddy/nav-dropdown.html
@@ -15,356 +15,356 @@
             </a>
             <div class="d-lg-none d-md-none d-flex me-auto p-0 ms-2">
                 <div>
-                <a class="text-body-secondary"
-                   href="{% url 'dashboard:dashboard' %}"
-                   aria-expanded="false"><i class="icon-2x icon-dashboard" aria-hidden="true"></i>
-            </a>
-            &nbsp;
-        </div>
-        <div>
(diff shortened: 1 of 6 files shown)


cdubz:
Closing this on favor of #1080

PR #1089, opened 2026-07-15
feat: notify user and disable actions when timer is ended in another session

### LINKED ISSUE #983: Feature request: Indicate timer already stopped
Use case: My wife is feeding, and sometimes she starts the timer and sometimes I do. Same for stopping the timer. However, sometimes it's not immediately clear if the timer was indeed stopped, because we might forget. This can lead to the following user-race-condition:

1. Timer is active
2. User A stops the timer (e.g. by indicating the feeding) details; user B didn't notice
3. User B was actively looking at the timer screen, believing the timer is still running
4. User B now presses "Feeding". This indeed goes to "Add a feeding", with a zero-length "now" feeding

Desired behavior: In step 3, the timer should now be greyed out, saying "timer already stopped".

EwertonG:
## Description
This PR addresses the issue of active timers getting out of sync when closed from a different device or browser session. 

When a timer is ended elsewhere, the detail page now detects the missing/inactive state (handling the API 404 response), stops updating the UI clock, displays a clear notification message to the user, and disables any invalid actions.

image

## Related Issues
- Closes #983

## Changes
- [x] Added periodic check to verify if the active timer still exists on the server.
- [x] Implemented handling for API `404 Not Found` errors to stop the countdown.
- [x] Added a warning message informing the user that the timer has already been ended.
- [x] Disabled timer-related action buttons once the end state is detected.

## Test Plan
- Verify that a 404 response on the timer poll endpoint triggers the alert banner.
- Ensure buttons (like stop, edit, or delete) are disabled after the notification appears.

### DIFF
--- a/babybuddy/templates/babybuddy/nav-dropdown.html
+++ b/babybuddy/templates/babybuddy/nav-dropdown.html
@@ -15,356 +15,356 @@
             </a>
             <div class="d-lg-none d-md-none d-flex me-auto p-0 ms-2">
                 <div>
-                <a class="text-body-secondary"
-                   href="{% url 'dashboard:dashboard' %}"
-                   aria-expanded="false"><i class="icon-2x icon-dashboard" aria-hidden="true"></i>
-            </a>
-            &nbsp;
-        </div>
-        <div>
-        <a class="text-body-secondary"
-           href="{% url 'core:timeline' %}"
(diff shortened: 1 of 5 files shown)


cdubz:
@EwertonG thanks for opening the PR! Could you get this caught with the latest changes to the main branch and resolve conflicts?

PR #1099, opened 2026-07-26
chore: reformat templates to satisfy djlint 1.42.x

lntutor:
## Problem

The `lint` gulp task runs `djlint --check .`. djlint is pinned as `djlint = "~=1.0"` in the `Pipfile` with no committed lockfile, so CI installs the latest 1.x at run time (currently **1.42.3**). A recent 1.x release changed template indentation rules, so `djlint --check .` now flags **13 existing templates** as needing reformatting — this is true on `master` itself, so the `test` job fails on **every open PR** regardless of its contents.

## Fix

Runs `djlint --reformat` (formatting only):
- 12 files are pure whitespace/indentation.
- 1 file (`widget_tag_editor.html`) has a hidden `<input>` collapsed onto one line — semantically identical.
- No Python changed; `prettier` already ignores `*.html` (per `.prettierignore`), so there's no formatter conflict.

After this, `djlint --check .` exits 0.

## Alternative

If you'd prefer to freeze formatting rather than chase djlint releases, pinning djlint to the previously-used version (and/or committing `Pipfile.lock`) is an equally valid fix — happy to switch this PR to that approach instead.

### DIFF
--- a/babybuddy/templates/babybuddy/form_field_no_label.html
+++ b/babybuddy/templates/babybuddy/form_field_no_label.html
@@ -31,5 +31,5 @@
     </div>
 {% endif %}
 {% if field.errors %}
-    <div class="invalid-feedback">{% for error in  field.errors %}{{ error }}{% endfor %}</div>
+    <div class="invalid-feedback">{% for error in field.errors %}{{ error }}{% endfor %}</div>
 {% endif %}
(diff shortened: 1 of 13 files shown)


cdubz (approved):

cdubz:
Thanks for the PR!

PR #1104, opened 2026-08-03
Allow django-imagekit >=6.0 (fix 500 on Python 3.14)

### LINKED ISSUE #1103: 500 error on /dashboard/ after updating to a build using Python 3.14
### Describe the bug

After updating my Docker image this morning, logging in redirects me to `/dashboard/` and I get a 500 Server Error. Every other page I've tried works fine. Turning on debug gives the traceback below, which points at django-imagekit and the child thumbnail template.

I noticed the new build moved to Python 3.14 (the previous one was on 3.13), and going back to the older image tag fixes it, so I suspect it's a Python version compatibility problem rather than anything in Baby Buddy itself. Posting here in case it's the right place, happy to move it if not.

### To Reproduce

1. Run Baby Buddy on a build using Python 3.14.
2. Have a child with a profile picture uploaded.
3. Log in and go to `/dashboard/`.
4. 500 Server Error.

It seems to depend on the child having a picture. I only worked that out from the traceback pointing at `child_thumbnail.html`, so I haven't tested removing the picture to confirm.

### Traceback

```
Request Method: GET
Request URL: http://<host>:8000/dashboard/

Django Version: 5.2.14
Python Version: 3.14.5
Exception Type: TypeError
Exception Value: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'
Exception Location: /lsiopy/lib/python3.14/site-packages/imagekit/hashers.py, line 22, in save_dict
Raised during: dashboard.views.Dashboard

Error during template rendering
In template /app/www/public/core/templates/core/child_thumbnail.html, error at line 3

  1 {% load imagekit static %}
  2 {% if child.picture %}
  3     {% thumbnail '40x40' child.picture as thumb %}
  4         <img src="{{ thumb.url }}" class="img-fluid rounded-circle" />
  5     {% else %}

  File "/lsiopy/lib/python3.14/site-packages/imagekit/hashers.py", line 22, in save_dict
    self._batch_setitems(sorted(obj.items()))

when serializing pilkit.processors.resize.Thumbnail state
when serializing pilkit.processors.resize.Thumbnail object
when serializing list item 0
when serializing list item 1
```

### Environment

- Baby Buddy version: v2.9.2
- Deployment: LinuxServer.io Docker image, tag `v2.9.2-ls243` (build date 2026-08-02)
- Python: 3.14.5
- Django: 5.2.14
- django-imagekit
- Database: SQLite (default)
- Host OS: Fedora Server, Docker

### Workaround

Pinning back to the previous image tag, `v2.9.2-ls242`, fixed it for me. That one is on Python 3.13. Both tags are the same Baby Buddy version, so nothing else changed and my data was unaffected.

playcations:
## Summary

Bumps `django-imagekit` from `~=5.0` to `~=6.0` (and pins `6.1.0` in `requirements.txt`) so Baby Buddy runs on **Python 3.14**.

## Problem

On Python 3.14, every page that renders a child thumbnail returns HTTP 500:

```
Django Version: 5.2.x
Python Version: 3.14
Exception Type: TypeError
Exception Value: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'
Exception Location: .../imagekit/hashers.py, line 22, in save_dict
Error during template rendering: core/templates/core/child_thumbnail.html
    {% thumbnail '40x40' child.picture as thumb %}
```

Python 3.14 changed the private pickle API to `_batch_setitems(self, items, obj)`. `django-imagekit` 5.0.0 calls it with a single argument, so `CanonicalizingPickler` (used to hash image specs) crashes whenever a thumbnail is generated.

This was fixed in **django-imagekit 6.0** ([matthewwithanm/django-imagekit#583](https://github.com/matthewwithanm/django-imagekit/pull/583), "Support Python 3.14"), but the current `~=5.0` constraint (`>=5.0,<6.0`) excludes it. It surfaced now that the LinuxServer.io image moved its base to Python 3.14. Reported in #1103.

## Change

- `Pipfile`: `django-imagekit = "~=5.0"` → `"~=6.0"`
- `requirements.txt`: `django-imagekit==5.0.0` → `django-imagekit==6.1.0`

`django-imagekit` 6.1.0 supports Python 3.9–3.14 and requires only `django-appconf` and `pilkit`, both already pinned (`pilkit==3.0` is unchanged), so no other dependencies change.

## Verification

- `django-imagekit` 6.1.0's `hashers.py` contains the version-conditional fix (`if sys.version_info[:2] >= (3, 14): self._batch_setitems(sorted(obj.items()), obj)`).
- Confirmed on a live Python 3.14 deployment (LinuxServer `v2.9.2-ls243`): applying the equivalent fix makes every previously-500ing authenticated page return 200, including `/dashboard/` with a child that has a profile picture.

Fixes #1103.


playcations:
Heads up on the failing `test` jobs — they're unrelated to this change. The failure is in `gulp lint`: `djlint` (unpinned at `~=1.0`, currently resolving to 1.43.2) wants to reformat 13 templates that this PR doesn't touch, so it's red on `master` for any PR right now. `black` and `prettier` pass, and the dependency change itself locks and installs cleanly (`django-imagekit==6.1.0`) — the run just never reaches the test step because lint fails first. Happy to adjust the constraint if you'd prefer a different pin.

playcations:
I also want to note that #1099 should fix testing here. @cdubz

MordechaiN:
Wanted to add independent confirmation from a production deployment, since this looks ready to merge and it may help to have another data point.

**Environment**
- LinuxServer.io `babybuddy` image, build `v2.9.2-ls243` (2026-08-02)
- Python 3.14.5
- django-imagekit 5.0.0 (as shipped)
- Django 5.2.14, pilkit 3.0

**Reproduction**
Independently reproduced the same exception on both `GET /feedings/add/` and `GET /feedings/<id>/` for any child with a profile picture set:

```
TypeError: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'
```

raised from `imagekit/hashers.py`'s `CanonicalizingPickler.save_dict()` during `{% thumbnail %}` cache-name hashing — matches the exception already discussed here.

Traced the trigger to `core/widgets.py`'s `ChildRadioSelect` (`option_template_name = "core/child_radio_option.html"` → `core/child_thumbnail.html` → `{% thumbnail '40x40' child.picture %}`). Confirmed via `core/forms.py` that this widget is shared across essentially every Add/Edit form with a `child` field (Feeding, Sleep, Diaper Change, Tummy Time, etc.) — so this isn't Feeding-specific, though our runtime testing (HTTP + `RequestFactory`) focused on the Feeding views specifically.

**Fix validation**
Upgrading only `django-imagekit` to the exact `6.1.0` release resolved the issue completely, with no BabyBuddy source changes required. `pip install --dry-run` showed no other dependency (Django, pilkit, django-appconf, Pillow) needed to move.

Validated post-upgrade:
- Existing feedings (multiple IDs) and newly created feedings both render correctly
- `/feedings/add/` and `/feedings/<id>/` return HTTP 200 for authenticated requests (previously 500)
- Confirmed via both live HTTP requests and read-only Django `RequestFactory` checks run inside the container
- No new errors in application logs after the upgrade
- The BabyBuddy REST API our Home Assistant integration talks to kept responding normally throughout, with no regressions observed

**Interim mitigation**
Since this fix hasn't reached an official image yet, we're running a temporary LinuxServer `custom-cont-init.d` script that installs `django-imagekit==6.1.0` at container start (only if the installed version is `<6.0`, so it becomes a no-op automatically the moment an official image ships the fix). Happy to remove it the day this lands in a release.

Thanks for putting this together — please let us know if it'd help to test a release candidate before this merges.

cdubz (approved):

cdubz (approved):

Issue #1108, opened 2026-08-07
PWA will not install behind Cloudflare Zero Trust


btoconnor:
The manifest file fails to load for me behind Cloudflare Zero Trust.  I am 99% certain that this is a result that was recently-ish (past few years) added that would require using [`crossorigin: use-credentials`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/crossorigin#use-credentials).  

This simply says that the manifest file exists behind authentication, and the browser needs to include the session cookies when fetching the file.  By default, it does not include the session cookies, which means that Cloudflare blocks the request. As far as I am aware, there are no real downsides to including this configuration option for those that are not behind a proxy.

I haven't had a chance to confirm this in a local development environment yet (my second child was born this week, so trying to find time when I can) - but as soon as I can confirm this I'm happy to submit the patch to fix it, as long as that's the desired route.

cdubz:
Congrats on #2! 😁 👶

btoconnor:
Confirmed locally. Will have patch momentarily.

cdubz:
Fixed by #1109

PR #1109, opened 2026-08-07
Set crossorigin to use-credentials for manifest file

btoconnor:
Addresses #1108 - sets `crossorigin` to `use-credentials`, which will instruct the browser to utilize the session cookies when fetching the manifest file, which will allow it to retrieve the file behind an authentication layer.


cdubz (approved):
Thanks, @btoconnor. Good to have you “back” 😁

btoconnor:
Good to be back! I don't use Alexa devices anymore so now I'll have to find something else to poke at around here 🙃

btoconnor:
@cdubz obviously no rush - but any idea when this will get cut into a new release?

PR #1113, opened 2026-08-14
Allow deselecting radio pill buttons by clicking again (#870)

### LINKED ISSUE #870: Diaper color unable to unselect
After creating a diaper change, it's possible to unselect the contents (wet or solid) after submitting by editing again. 
This might be fine and useful if one decides afterwards to not specify the contents.

What's not working is, after choosing a color and submission, to edit again and unselect the color. 
Happened a few times, that it was chosen by mistake even though it was wet content. 
So it was necessary to delete and submit a new diaper change after all in order to have no color chosen/submitted.

SoTHISIsFitness:
## Description

Allows users to deselect a previously selected radio pill button by clicking it again. This addresses #870.

Currently, once a radio button (pill style) is selected, there is no way to unselect it without choosing a different option or reloading the page. This is especially problematic for fields that are optional, such as diaper change solid waste color tracking.

## Changes

Adds a small vanilla JavaScript snippet (no dependencies) to two template partials:

- `core/templates/core/pill_radio.html` — generic pill radio widget
- `core/templates/core/child_radio.html` — child selector pill widget

FYI: allowing for the child selector pill to be de-select-able is intentional, and is in support of other feature development work I've got going on that I haven't yet submitted PR's for (but it'll come).

The script uses event delegation on the container:

1. On click, finds the target label and its associated radio input
2. If the input is already checked, prevents the default selection behavior
3. Unchecks the input and dispatches a `change` event so any listeners (e.g., form auto-submit) fire correctly

Uses a `dataset.deselectInit` guard to prevent double-binding when multiple pill groups are on the same page.

## Testing

- Verified on Chrome, Firefox, and Safari
- Tested with child selector (single child + household scope) and diaper color fields
- Confirmed `change` event fires correctly for dependent UI
- No console errors

Closes #870


SoTHISIsFitness:
@cdubz - nudging to see if we can make some moves on some of these pending PR's?

cdubz (changes requested):
@SoTHISIsFitness I don't like this approach -- it feels like it's fighting the default behavior of the element too much.

In theory, I think we can just add an empty choice option to the models to support a deselect? For color anyway.

> FYI: allowing for the child selector pill to be de-select-able is intentional, and is in support of other feature development work I've got going on that I haven't yet submitted PR's for (but it'll come).

Let's not include that in this PR.

PR #1116, opened 2026-08-14
Link back-to-back feedings as continuations (#837)

### LINKED ISSUE #837: Group multiple feeding types in one feeding
### Issue

We're pumping breast milk, and supplementing with formula when there isn't enough breast milk. So most of our feedings consist of both breast milk and formula, both by bottle.

Right now, I need to log these as 2 separate adjacent feedings so the amounts of formula vs breast milk is correct. However, this means that the feeding frequency graph is useless to me, since the interval between every other feeding is of zero duration.

image

### Request
I'd like a way to group feedings so I can identify a formula and breast milk feed as a single feeding session.

### Possible solution
Create a new data type, "feeding session", that can have multiple feedings associated with it, and base feeding intervals off of that new data type

SoTHISIsFitness:
## Description

Adds the ability to link feedings that are part of the same feeding session. Addresses #837.

When a feeding session spans multiple entries (e.g., breast milk followed by formula supplement, or a feeding paused and resumed), these can now be explicitly linked so they're identifiable as one logical session.

## Changes

**Feeding model** — new optional self-referencing FK:
- `previous_feeding` — links to the feeding this one continues
- `on_delete=SET_NULL` so deleting the earlier feeding doesn't cascade
- `related_name="continuations"` exposes the reverse relationship

**Auto-linking** in `FeedingForm`:
- When adding a new feeding, if the most recent feeding ended within the configured threshold (default 30 minutes) of this one's start, `previous_feeding` is auto-selected — the user can override or clear it
- The threshold is site-configurable via the existing django-dbsettings pattern: `FeedingSettings.continuation_threshold_minutes` (Settings → Feeding), alongside the existing `feeding_diff_end` setting
- When editing an existing feeding, the dropdown shows feedings within +/-6h of this feeding's start time — the relevant window — rather than only recent feedings, so older entries stay editable

**API**: `previous_feeding` field exposed in the serializer; API test updated.

## Usage

1. Log the first part of the feeding normally
2. When logging the continuation, the previous feeding is pre-selected in the "Previous feeding" dropdown
3. Or manually pick the feeding this entry continues

This keeps amounts per-type accurate (solving the mixed-feeding logging problem from #837) while making the session grouping explicit.

Closes #837


cdubz:
I'm a little hesitant about the auto-linking behavior here. I'd be more open to adding the underpinnings and a _manual_ linking flow separately and then considering atuo-link as a separate PR. Thoughts?

SoTHISIsFitness:
Manual linking is actually already a part of this PR - and I actually just realized right now that only one part of the intended effect of "continuation" is happening: showing it on my "Recent Consumption" dashboard card. The other intended effect is adjusting what data is shown in the "last feeding" dashboard card (and similar uses of that data point, present or future). So that's something for me to update anyway.

With the auto-linking being based on a user- configurable time between feedings, I figured it should be flexible enough that anyone could set a time window they're comfortable with - I'm sure everyone has a time threshold where if a feeding starts within x minutes of the previous one... it's definitely a continuation. 

That said... we could also leave both manual & auto in the PR and just include an "enable / disable" toggle for the auto (along with the configurable time threshold) and have it be default-off unless the user enables it? I need to update the feature before you merge it anyway, so that could be a good solution.

cdubz:
> With the auto-linking being based on a user- configurable time between feedings, I figured it should be flexible enough that anyone could set a time window they're comfortable with - I'm sure everyone has a time threshold where if a feeding starts within x minutes of the previous one... it's definitely a continuation.

This does raise a question... does it make sense as a user setting as opposed to a site setting? The app is really designed primarily for "single family" use. In that way, I think it may be confusing if caregiver A has the user setting, makes an entry, then (for whatever reason) caregiver B (or has no setting or a different) makes the second entry.

> That said... we could also leave both manual & auto in the PR and just include an "enable / disable" toggle for the auto (along with the configurable time threshold) and have it be default-off unless the user enables it? I need to update the feature before you merge it anyway, so that could be a good solution.

I agree that manual and auto makes sense. From an implementation perspective, we should start with just manual. Happy to review an auto function PR as well, but I'd want more community input on that.

PR #1117, opened 2026-08-14
Faster UI and SPA-feel navigation with Turbo Drive

Helvio88:
## Problem

Full page reloads and a heavy first-load JavaScript bundle make the UI feel slow. Every page currently loads vendor/app scripts immediately, includes Masonry everywhere, and ships every Plotly locale. Timeline and dashboard views also do extra database work (N+1 child/tag queries, full model rows for statistics).

## Changes

No visual or functional change is intended. This PR keeps the same screens and features, and makes navigation and first paint faster:

- Defer vendor/app JavaScript so first paint is not blocked. Load Masonry only on the child dashboard.
- Load Plotly only on report pages, and only the current language locale instead of bundling every Plotly locale.
- Use `select_related`/`prefetch_related` on the timeline, and compute dashboard statistics with aggregates/`values_list` instead of loading full model rows.
- Add Hotwire Turbo Drive for in-app navigation, with prefetch on hover/touch. Login, logout, file uploads, the API browser, and Django admin still do a full reload (`data-turbo="false"`). Page JavaScript re-inits on `turbo:load`.
- Tag chips are now buttons (`type="button"`) so mobile Safari taps work under Turbo. No visual change.

## Tests

Tests were added or updated for Turbo Drive, login isolation, timeline query behavior, dashboard stats, report Plotly loading, and tag-editor button markup.

`python manage.py test --settings=babybuddy.settings.test --parallel --exclude-tag isolate` (335 tests) and the isolated `ViewsTestCase.test_password_reset` both passed locally.

## Static assets

Production `/static` assets were rebuilt and committed after the `static_src` changes.

## Docker

This PR does not add a Dockerfile. Docker deployments remain the LinuxServer.io image documented in the existing setup guide.


Helvio88:
Follow-up: tag chips are now buttons so mobile Safari taps work under Turbo. No visual change.

PR #1118, opened 2026-08-15
Configurable wording for "Previous" column time gaps

### LINKED ISSUE #951: In the English localization, use of the word "ago" on events older than the most recent is incorrect
I realize this is barely an issue, but "ago" is always used for measures of time compared to the current time. On timelines for feedings, changings, pumping, sleep, and tummy time the events prior to the most recent display the time between it and the previous event, so "ago" is not appropriate. Perhaps use "before" in its place, or even nothing at all since the field is already named "previous" and the text could be considered redundant.

I'll see about setting up an environment to edit the translation myself (if it is entirely that easy in this case) and submit a PR if this even sounds like something worth caring about.

SoTHISIsFitness:
## Description

The activity list pages (feedings, changes, pumpings, sleep) render the time gap to the previous entry as a fixed "X minutes **ago**" in the Previous column. For users (or locales) where "ago" reads oddly, this makes the trailing word configurable.

Small standalone enhancement — no upstream issue that I'm aware of; happy to retarget if there's a matching one.

## Changes

- New `DisplaySettings` dbsettings group with a single `prev_word` choice value (`ago` / `prior` / `before`, default `ago`), editable on the existing `/settings/` site settings page
- New `{% prev_word %}` template tag (`core/templatetags/misc.py`); uses the `{% prev_word as w %}` + `{% blocktrans %}` pattern because `blocktrans` can't contain arbitrary tags inline
- Wired into all four list templates — feeding, diaper change, pumping, sleep — replacing the hardcoded `ago`
- The settings group is instantiated at module import so it registers with dbsettings eagerly (unattached groups only register on first access, which would hide the value on the settings page in a fresh process)

## Testing

- Verified the tag end-to-end in a container: default `ago`, changed to `prior` and `before` via the dbsettings API, reflected immediately in rendered templates
- Confirmed the value appears with its label ("Previous-entry wording") on `/settings/` in a fresh process, before any template using the tag renders
- Full test suite: identical results to an unmodified `master` baseline in the same harness (no new failures)

## Notes

- No new tests included — the behavior is one tag + four one-line template swaps; happy to add a `prev_word` tag test if maintainers want one
- would close #951 (I'm pretty sure, anyway)


cdubz:
> The activity list pages (feedings, changes, pumpings, sleep) render the time gap to the previous entry as a fixed "X minutes ago" in the Previous column. For users (or locales) where "ago" reads oddly, this makes the trailing word configurable.

Is this not better handled in l10n? I understand that individual users may have differing opinions about what word makes the most here but this feels excessive.

> would close https://github.com/babybuddy/babybuddy/issues/951 (I'm pretty sure, anyway)

I don't think this does close #951 -- that is about the context of the word, not necessary the word itself.

PR #1119, opened 2026-08-17
feat(core): add autocomplete of prior names to the medication form

AnwariasEu:
Add an HTML5 <datalist> of previously-used medication names to the Medication add/edit form's Name field. The field stays free text — a brand-new name is accepted and becomes a future suggestion — while prior names are offered as a dropdown. Suggestions are global across all children and ordered by most-recent use.

Implemented as a reusable DatalistTextInput widget wired up in MedicationForm; no model change and no migration.

**Rational:** I needed to give the same medication every day and was tired of always typing in the same name very time, trying to 100% match the previous names.

**Potential performance impact:** core_medication is queried in full on every medication add/edit page load. Returned results are not limited (if you see this as a problem a generous limit 50 would make a reasonable global limit). Both should not be an issue for a few thousand medication entries.

**Disclaimer:** Code changes created by Claude Code. Reviewed by a human (me). Tested for multiple days now on my own productive instance.

**How it looks:**
image


cdubz (changes requested):
@AnwariasEu this looks like a nice feature, thanks for the PR! Please take a look at and address the items Copilot review flagged.

PR #1123, opened 2026-08-26
fix(timeline): show dose wore off event on correct day (closes #1059)

### LINKED ISSUE #1059: Inconsistent display behavior for timeline events that cross days

aniruddhaadak80:
## What Problem This Solves

When a medication is taken late in the day and the next dose is due the following day, the "dose wore off" timeline event appeared on the day the medication was taken instead of the day the next dose is actually due. This was confusing — viewing the timeline for the 23rd showed an empty day while the 22nd showed a "dose wore off" event at 04:31 belonging to the 23rd.

## Why This Change Was Made

The `_add_medication` function queried medications taken on the current day and unconditionally appended a "dose wore off" event at `next_dose_time`, regardless of whether that time fell on a different day. Additionally, medications taken on previous days whose `next_dose_time` falls today were never shown on today's timeline.

## What Changed

- The "dose wore off" event is now only appended when `next_dose_time` falls within the viewed day's date range (`min_date <= next_dose_time <= max_date`)
- Added a second query to find medications taken on previous days whose `next_dose_time` falls within the current day, so those "dose wore off" events appear on the correct day

## How Tested

- Existing `MedicationTestCase` tests pass (5/5)
- Manually verified the timeline logic handles same-day and cross-day scenarios correctly
- The fix is bounded to the `_add_medication` function with no side effects on other timeline event types

Fixes #1059


aniruddhaadak80:
Diagnosis: all 5 'test' matrix jobs failed in the 'gulp lint' step because black (26.5.1) wanted 'core/timeline.py' line 290 collapsed to a single line ('1 file would be reformatted'). This line came from this PR's new medication block. Fixed in 2a8fae5; no behavior change.

aniruddhaadak80:
Diagnosis of the red `test (3.10–3.14)` jobs (run 34026621403, `gulp lint` step): not caused by this PR's diff.

Breakdown of the `lint` task in that run: Prettier passes ("All matched files use Prettier code style!"), Black passes ("192 files would be left unchanged" — includes `core/timeline.py` after `2a8fae5`), but `djlint --check .` fails with 50+ "Formatting changes required" errors. This PR touches only `core/timeline.py`, which djlint does not lint, so those failures are pre-existing template drift flagged repo-wide by the pinned `djlint==1.45.2` — independent of this change.

Reformatting dozens of unrelated templates is out of scope here; this needs a maintainer call (run `gulp format` on the base branch or adjust the djlint pin/config). Leaving the diff as-is.

cdubz (changes requested):
@aniruddhaadak80 thanks for the PR! This needs to be updated (if it's still required) based on changed from #1080.

PR #1124, opened 2026-08-27
feat: add 'Keep screen on' toggle to the timer page

### LINKED ISSUE #1070: Add "keep screen on" toggle to timer page to prevent device sleep
## Feature Request

Add a toggle/switch on the timer page that allows the user to keep the screen on and prevent the phone from being locked automatically by the OS. The toggle state should persist in session storage so that if a user enables it, it will remain enabled when returning to the timer page during the same browser session.

### Suggested Implementation
- Location: Timer page UI, clearly visible toggle (e.g., checkbox or switch), label: "Keep screen on"
- State is read from and written to `sessionStorage`, so the user's setting is preserved while the browser is open
- When enabled, use the [Screen Wake Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API) when supported, preventing the device from sleeping
- The toggle is restored to its last state on every page visit (from sessionStorage)
- Wake Lock is released when navigating away from the page or disabling the toggle
- For privacy and battery, default should be off on a new browser session
- Handles re-acquiring wake lock on visibility changes (e.g., switching tabs)
- Provides graceful fallback for browsers that do not support the Wake Lock API

### Example UI (timer_detail.html):
```html
<div class="form-check form-switch d-inline-flex align-items-center">
    <input class="form-check-input" type="checkbox" id="keepScreenOnToggle">
    <label class="form-check-label ms-2" for="keepScreenOnToggle">Keep screen on</label>
</div>
```

### Example JavaScript (screen-lock.js):
```javascript
BabyBuddy.ScreenLock = (function () {
  var wakeLock = null;
  const STORAGE_KEY = "babybuddy_keep_screen_on";

  var ScreenLock = {
    init: function () {
      var toggle = document.getElementById("keepScreenOnToggle");
      if (!toggle) return false;
      var savedState = sessionStorage.getItem(STORAGE_KEY);
      if (savedState === "true") {
        toggle.checked = true;
        ScreenLock.acquireWakeLock();
      }
      toggle.addEventListener("change", function () {
        if (this.checked) {
          ScreenLock.acquireWakeLock();
        } else {
          ScreenLock.releaseWakeLock();
        }
      });
      document.addEventListener("visibilitychange", function () {
        if (!document.hidden && wakeLock === null && toggle.checked) {
          ScreenLock.acquireWakeLock();
        }
      });
      window.addEventListener("beforeunload", function () {
        ScreenLock.releaseWakeLock();
      });
      return true;
    },
    acquireWakeLock: async function () {
      try {
        if ("wakeLock" in navigator) {
          wakeLock = await navigator.wakeLock.request("screen");
          sessionStorage.setItem(STORAGE_KEY, "true");
        } else {
          console.warn("Wake Lock API not supported.");
        }
      } catch (err) {
        console.error("Failed to acquire wake lock", err);
      }
    },
    releaseWakeLock: async function () {
      try {
        if (wakeLock !== null) {
          await wakeLock.release();
          wakeLock = null;
          sessionStorage.setItem(STORAGE_KEY, "false");
        }
      } catch (err) {
        console.error("Failed to release wake lock", err);
      }
    },
  };
  return ScreenLock;
})();
```

#### Notes
- See [MDN Web Docs: Screen Wake Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API)
- Do not include "timer" in the toggle label unless additional context is needed
- Only persistent for the session (sessionStorage)
- For full implementation, initialize with: `BabyBuddy.ScreenLock.init();` in a script tag on the timer page.

---
**Labels:** `enhancement`

Agi-Asi:
### Feature

Adds a visible "Keep screen on" switch to the timer detail page, implementing #1070 with exactly the semantics proposed there:

- Uses the [Screen Wake Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API) to prevent the device from sleeping while a timer is running.
- Toggle state persists in `sessionStorage`, so it's restored on every timer page during the same browser session.
- The lock is **re-acquired automatically** when the page becomes visible again — browsers release wake locks on tab switch/minimize, so without this the toggle would silently stop working after the first app switch.
- The lock is released when the toggle is turned off; page navigation releases it implicitly per the API spec.
- If the browser refuses the request (e.g. battery-saver mode), the toggle flips back off and the stored state is cleared — the UI reflects reality instead of pretending the screen will stay on.
- On unsupported browsers / insecure contexts the whole control is hidden rather than offering a switch that cannot work.

### Implementation notes

- No new dependencies — a small `BabyBuddy.WakeLock` module in the existing `core/static_src/js/timer.js`, plain DOM APIs.
- Bootstrap `form-switch` markup on the timer detail template; label wrapped in `{% trans %}` for translation.
- This complements (doesn't replace) the existing focus-refresh in `BabyBuddy.Timer`, which corrects drift after the screen *did* sleep.

### Testing

- `gulp scripts` builds cleanly; `prettier --check` passes on the JS; `djlint --check` passes on the template; `black --check .` unaffected.
- Manually traced the four state paths (enable→acquire, refuse→uncheck, hide on unsupported, visibilitychange re-acquire); the module is defensive about a missing checkbox element.

Closes #1070


cdubz (changes requested):
@Agi-Asi could you take a look at the notes from the Copilot review? I caught this up to `master` and put the static assets in but I think there are some other valid points there.

PR #1133, opened 2026-09-05
fix(reports): count whole days in diaper lifetimes graph

NotAFlightRisk:
The Diaper Lifetimes chart wraps every gap at 24 hours. A 30 hour gap between changes plots as a 6 hour point, and a gap of exactly 24 hours doesnt show up at all. `timedelta.seconds` is only the sub-day remainder, so dividing by 3600 throws the days away, and the `> 0` guard reads a clean 24 hours as zero and skips it.

You need a day where you didnt log anything, which most people manage sooner or later. `diaperchange_intervals.py` and `feeding_intervals.py` already do the same sum with `total_seconds()` in both spots, so this just brings the third one in line.

No test in here on purpose: #1074 adds `reports/tests/graphs/tests_diaperchange_lifetimes.py` already, and a second copy of that file would only fight with it.

---

Separate thing: this is the only graph in `reports/graphs` that pins `yaxis.dtick` to 1. Was fine while every value sat under 24, bit less so now one can be a few hundred.


cdubz (approved):
@NotAFlightRisk Thanks for the catch and fix!

I'll try to get to #1074 soon too...

PR #1136, opened 2026-09-09
fix: render the 404 page instead of raising a template error

Daltonganger:
## The bug

`babybuddy/templates/error/404.html` line 9 reads:

```django
{% blocktrans trimmed with path="<code>"|add:request_path|add"</code>"|safe %}
```

The second `add` is missing its colon, so Django parses it as a filter with
no argument and raises while compiling the template:

```
django.template.exceptions.TemplateSyntaxError: add requires 2 arguments, 1 provided
```

The 404 handler only renders this template when `DEBUG` is off, which is every
deployment, so **any missing path returns a 500 instead of a 404** in
production. With `DEBUG` on, Django uses its own debug 404 page and the
template is never touched, which is why this has gone unnoticed.

## The fix

One character:

```diff
-{% blocktrans trimmed with path="<code>"|add:request_path|add"</code>"|safe %}
+{% blocktrans trimmed with path="<code>"|add:request_path|add:"</code>"|safe %}
```

## Test

No existing test rendered the template, for the same reason the bug survived:
nothing runs the handler with `DEBUG` off. `ErrorPageTestCase` uses
`@override_settings(DEBUG=False)`, requests a path that does not exist, and
asserts a 404 whose body contains both the heading and the requested path.

Verified it is a real regression test: with the colon removed again it errors,
with the fix it passes.

## Notes

`gulp lint` is currently failing on `master` and on every open pull request,
including this one, at `djlint --check`. `Pipfile` has `djlint = "~=1.0"` with
no lockfile, so CI resolves djlint 1.46.1, while the committed templates are
formatted for 1.44.2. That is unrelated to this change — this branch produces
the same 118 findings as `master`, and zero under 1.44.2.


cdubz (approved):
@Daltonganger thanks for the find and fix!

Issue #985, opened 2025-06-16
/api/schema throws 500 error


MaZZly:
Trying to access the DRF live schema throws 500 error.

Error shown when DEBUG:True
```
Environment:


Request Method: GET
Request URL: https://foo.bar.baz/api/schema

Django Version: 5.1.6
Python Version: 3.12.10
Installed Applications:
['api',
 'babybuddy.apps.BabyBuddyConfig',
 'core.apps.CoreConfig',
 'corsheaders',
 'dashboard',
 'reports',
 'axes',
 'django_filters',
 'rest_framework',
 'rest_framework.authtoken',
 'widget_tweaks',
 'imagekit',
 'storages',
 'import_export',
 'qr_code',
 'dbsettings',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize']
Installed Middleware:
['babybuddy.middleware.HomeAssistant',
 'django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'babybuddy.middleware.RollingSessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'babybuddy.middleware.UserTimezoneMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'babybuddy.middleware.UserLanguageMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'axes.middleware.AxesMiddleware']



Traceback (most recent call last):
  File "/lsiopy/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/lsiopy/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1140, opened 2026-09-12
fix: restore OpenAPI schema generation

Daltonganger:
## Summary

`django-filter` 25.1 removed the schema generation methods of its DRF filter
backend in favour of `drf-spectacular`
([CHANGES.rst](https://github.com/carltongibson/django-filter/blob/main/CHANGES.rst)),
but DRF's OpenAPI schema generator still calls
`get_schema_operation_parameters()` on *every* configured filter backend:

```python
# rest_framework/schemas/openapi.py
for filter_backend in self.view.filter_backends:
    parameters += filter_backend().get_schema_operation_parameters(self.view)
```

Because `DjangoFilterBackend` no longer implements that method, both the
`/api/schema` endpoint and the `generateschema` management command fail:

```
AttributeError: 'DjangoFilterBackend' object has no attribute 'get_schema_operation_parameters'
```

This is the same failure reported in #985. As a side effect, `openapi-schema.yml`
could not be regenerated and has been stale since v2.10.0.

## Changes

- Add `api.filter_backends.FilterBackend`, a `DjangoFilterBackend` subclass that
  re-implements `get_schema_operation_parameters()`.
- Use it as `DEFAULT_FILTER_BACKENDS` in the base settings.
- Add regression tests for the schema endpoint and for the presence of filter
  parameters in the generated schema.

The generated schema is **unchanged**: filter parameters keep the same names,
descriptions and types. Verified by generating the schema with the restored
command and comparing it to the committed `openapi-schema.yml` — the two parse
to identical data structures. The method is adapted from django-filter 24.3,
which is BSD licensed like Baby Buddy.

## Why not `drf-spectacular`?

I tried it before writing the subclass, and it is a real option — but it is a
migration of the schema layer, not a like-for-like swap. Measured on this
codebase (drf-spectacular 0.30.0, with `DEFAULT_SCHEMA_CLASS` pointed at its
`AutoSchema`):

- All 92 `operationId`s change (`createFeeding` → `feedings_create`,
  `listChildren` → `children_list`, …). Anything that generates a client from
  `openapi-schema.yml` gets new method names across the board.
- Every response body moves from inline objects to a `components/schemas`
  block of 54 named schemas, so the diff of the committed file is close to a
  full rewrite.
- `api/views.py` sets `schema = AutoSchema(operation_id_base=...)` on
  `ProfileView`, which drf-spectacular rejects outright (`AssertionError:
  Incompatible AutoSchema`) until that view is changed too.
- One serializer warning remains: `ProfileSerializer.get_api_key` needs a type
  hint or `@extend_schema_field` before the schema is warning-free.
- The output version becomes OpenAPI 3.0.3 instead of 3.0.2.

None of that is a blocker, and if the project wants to adopt `drf-spectacular`
I am happy to do that migration as its own PR. It just seemed the wrong vehicle
for "the schema command has been broken since v2.10.0" — that is a regression
fix with a two-file footprint, and this keeps the published schema stable for
anyone consuming it. The subclass can be deleted wholesale once a real schema
framework lands.

## Test plan

- `python manage.py generateschema ... --file openapi-schema.yml` now runs and
  produces output identical to the committed file.
- Full suite green: 429 tests with `--exclude-tag isolate` on top of the
  current `master` (6dfc907e), which includes #1134.
- `black . --check`, `djlint --check .` and `prettier . --check` are clean.


cdubz (commented):
@Daltonganger is there a reason _not_ to just refactor to use `drf-spectacular`?

Daltonganger:
Good question — I tried it before writing the subclass, and it works, but it turns out to be a migration of the schema layer rather than a like-for-like swap. I measured it on this codebase (drf-spectacular 0.30.0) and added the full details to the PR description under "Why not `drf-spectacular`?". The short version:

- All 92 `operationId`s change (`createFeeding` → `feedings_create`, …), so anything generating a client from `openapi-schema.yml` gets new method names across the board.
- Every response body moves from inline objects to a `components/schemas` block of 54 named schemas — the committed schema file would be close to a full rewrite.
- `ProfileView` sets `schema = AutoSchema(operation_id_base=...)` explicitly, which drf-spectacular rejects with an `AssertionError` until the view is changed too, and `ProfileSerializer.get_api_key` needs a type hint or `@extend_schema_field` before generation is warning-free.
- Output becomes OpenAPI 3.0.3 instead of 3.0.2.

None of that blocks adoption, and if you would rather move to `drf-spectacular` I am glad to do that as its own PR — the subclass is small enough to delete wholesale afterwards. It just felt like the wrong vehicle for "the schema command has been broken since v2.10.0": that is a regression fix with a two-file footprint, and this way the published schema stays byte-identical for anyone already consuming it.

Both this and #1146 are rebased on current `master` (6dfc907e, includes #1134) and green: 429 tests, `black`/`djlint`/`prettier` clean, and the regenerated schema parses identical to the committed file.

cdubz (changes requested):
@Daltonganger OK that makes total sense -- on minor CR: can you add some of that context to the class documentation in code as well? We can always track back to this PR, but I think it will useful to also note it right in the docstring.

PR #1141, opened 2026-09-13
fix(core): populate the read-only group on first migrate

Daltonganger:
## Summary

The `read_only` group is empty on a freshly migrated database, so a read-only user can see nothing at all.

`core`'s `post_migrate` receiver was connected with `sender=self`, which fires once, on `core`'s own signal — before the permissions it needs exist:

- `core`'s permissions are created by `django.contrib.auth`'s `post_migrate` receiver, and `django.contrib.auth` is listed *after* `core` in `INSTALLED_APPS`, so on `core`'s own signal they are not in the database yet.

The receiver therefore finds no permissions and adds nothing. A fresh install ends up with an empty `read_only` group; only instances that ran a later migration (which fires the signal again, for another app) happened to converge.

## Changes

- `core/apps.py`: connect the receiver without a `sender` filter and with a `dispatch_uid`, so it runs on every application's `post_migrate` signal and converges to the complete set. `Group.permissions.add()` makes the repeats idempotent.
- `core/tests/tests_apps.py`: regression tests for a freshly migrated database.

## Test plan

- [x] `python manage.py test core.tests.tests_apps` — passes
- [x] The same tests **fail without this change** (`AssertionError: False is not true` on `user.has_perm("core.view_child")`), so the regression is genuinely covered
- [x] Full suite: 334 tests, same 6 pre-existing failures in `babybuddy.tests.tests_views` that are present on unmodified `master`
- [x] `black`, `djlint`, `prettier` clean

## Notes

This is a standalone bug fix. It was previously carried inside #1134; it has been split out at the maintainer's request so it can be reviewed and merged on its own.


cdubz (approved):
@Daltonganger looks good, thanks!

PR #1143, opened 2026-09-13
fix: apply model permissions to reports, the timeline and dashboard cards

Daltonganger:
## Summary

Several views only check `core.view_child` before rendering data from other models, so a user with permission for one kind of entry can read all of them.

This is a data exposure bug, independent of any user-role feature. It is most visible with a restricted account, but it affects every non-superuser: the reports, the timeline and parts of the dashboard render entries from models the user has no permission for.

Concretely, on unmodified `master`, a user with **only** `core.view_child`:

```
GET /children/<slug>/reports/medication/frequency/  ->  200
GET /children/<slug>/reports/medication/intervals/  ->  200
```

and after this change:

```
GET /children/<slug>/reports/medication/frequency/  ->  403
GET /children/<slug>/reports/medication/intervals/  ->  403
```

Measured with `manage.py test`, not inferred from reading the code.

## Changes

- `reports/views.py`: each report now requires `view` on the model it renders, on top of `core.view_child`. The report list only offers the reports the user may open.
- `core/timeline.py`: `get_objects()` takes the user and leaves out event types they have no `view` permission for. Passing `None` includes everything, preserving the existing callers' behaviour.
- `dashboard/templatetags/cards.py`: the cards that combine several models into one card check each model's permission.
- `dashboard/templates/dashboard/child.html`, `dashboard/templates/dashboard/dashboard.html`, `reports/templates/reports/report_list.html`: hide entries the user may not open.
- Tests for all three, including read-only cases so the existing behaviour stays covered.

## Test plan

- [x] `python manage.py test` — 340 tests, only the 6 pre-existing failures in `babybuddy.tests.tests_views` that are also present on unmodified `master`
- [x] The leak above was reproduced on unmodified `master` and closed by this change
- [x] `black`, `djlint`, `prettier` clean
- [x] Standalone: applies cleanly to `master`, and the permission tests do not depend on any other pending change

## Notes

Split out of #1134 at the maintainer's request, since it is a value fix rather than part of the role feature. Rebased onto current `master`.

`core/tests/tests_views.py` carries CRLF line endings on `master`; the change preserves them.


cdubz (approved):
Thanks, @Daltonganger! This is a great catch and fix.

PR #1144, opened 2026-09-13
fix: correct the permission codenames used by head circumference and tags

Daltonganger:
Django returns `False` from `has_perm()` and from the `{% if perms.x.y %}` template tag for a permission that does not exist, rather than raising. A codename that does not match the one Django creates therefore **denies access to every user who holds the correct permission**, silently, with nothing in the logs to explain it.

Two models are affected.

**Head circumference.** `HeadCircumference` declares `model_name = "head_circumference"`, but Django derives permission codenames from the model class name (`HeadCircumference` → `headcircumference`), not from `model_name`. The four permissions Django creates are `add_headcircumference`, `change_headcircumference`, `delete_headcircumference` and `view_headcircumference`. The views and templates asked for `*_head_circumference`, which never exists.

**Tags.** `Tag` generates `view_tag`, while both tag views asked for `view_tags`.

### Measured effect

A user holding exactly the four real head circumference permissions, and `core.view_tag`:

| URL | before | after |
| --- | --- | --- |
| `/head-circumference/` | **403** | 200 |
| `/head-circumference/add/` | **403** | 200 |
| `/tags/` | **403** | 200 |
| `/tags/add/` | 200 | 200 |

`/tags/add/` already worked because `TagAdminAdd` asks for `core.add_tag`, which is correct; only the two `view_tags` references were wrong.

### Why nothing caught it

The read-only group is built from `apps.all_models`, which yields the correct codenames, so `read_only` users were never affected. This only breaks a user or group that was granted permissions explicitly — for example through the admin permission picker, which is the documented way to give a caregiver more than its group. `grep` for the codename, `manage.py check` and the existing tests all pass either way, because a permission that does not exist is indistinguishable at those levels from one the user simply lacks.

### Changes

- `core/views.py` — four head circumference codenames plus the two `view_tags` references.
- `babybuddy/templates/babybuddy/nav-dropdown.html` — the two head circumference menu entries were hidden from every user.
- `core/templates/core/head_circumference_list.html` — the add/change/delete buttons were hidden from every user.
- `core/tests/tests_permissions.py` *(new)* — four tests.

The tests fail with the same `403 != 200` against the previous code, and additionally assert that **no view or template under `core` or `reports` refers to a `core.*` permission that does not exist**, so a future typo is caught even on a code path that no request exercises yet.

### Verification

- `core.tests.tests_permissions`: 4 failures against the previous code, 4 pass with the fix.
- Full suite: 335 tests, with only the 6 pre-existing `babybuddy.tests.tests_views` failures that are also present on unmodified `master` (`664ac751`). Those are unrelated to this change and are not touched here.
- `gulp lint` equivalent: `black` reports 194 files unchanged, `djlint` reports 0 files to update, `prettier` clean.


cdubz (approved):
Great catch -- thanks, @Daltonganger!

PR #1145, opened 2026-09-13
fix: stop the password reset test from logging out the shared client

Daltonganger:
`manage.py test` without the runner's flags reports six failures in `babybuddy.tests.tests_views`. They are not regressions and they are not flaky, and the cause is a single line in one test.

### The cause

`ViewsTestCase` logs in once in `setUpClass` and shares that client across every test in the class:

```python
cls.c = HttpClient()
cls.c.login(**cls.credentials)
```

`test_password_reset` then calls `self.c.logout()` on it. Django orders tests alphabetically within a class, so `test_password_reset` runs **before** the six tests that fail, and each of them inherits a logged-out client:

| test | failure |
| --- | --- |
| `test_rolling_sessions` | `'Set-Cookie: sessionid=""...' == 'Set-Cookie: sessionid=""...'` |
| `test_root_router` | `'/login/?next=/' != '/dashboard/'` |
| `test_user_settings` | `302 != 200` |
| `test_user_unlock` | `302 != 200` |
| `test_user_views` | `302 != 403` |
| `test_welcome` | `302 != 200` |

Every one of those is the same symptom: the request was redirected to the login page. Proven directly — running `test_password_reset` followed by `test_root_router` fails, and the same two tests in the reverse order pass.

### Why it went unnoticed

The test carries an `isolate` tag, which the runner turns into `--exclude-tag isolate` plus a separate invocation of the tagged test on its own. That works around the symptom and hides the cause: the suite passes under `gulp test`, and six tests fail under a plain `manage.py test`. The failures look exactly like real regressions.

The tag also has the effect of running `test_password_reset` outside the main suite entirely, so it was the only test in the class not covered by the run that everything else reports on.

### The fix

One line of intent: the test takes its own client rather than logging out the shared one.

```python
client = HttpClient()
```

This covers only this test. The task that introduced the tag (#148) was about running a test by itself; the coupling here is different — one test mutating state that seven other tests depend on — and a separate client fixes it at the source rather than excluding a test from the run.

### Verification

Four invocations, all green:

| invocation | before | after |
| --- | --- | --- |
| `--parallel --exclude-tag isolate` (runner) | 330 OK | 331 OK |
| `manage.py test`, no flags | **6 failures** | 331 OK |
| `ViewsTestCase` alone | **6 failures** | 10 OK |
| `test_password_reset` alone | 1 OK | 1 OK |

The suite count goes from 330 to 331 because the tagged test now runs as part of the main suite instead of separately.

`black` reports 193 files unchanged, `djlint` 0 files to update, `prettier` clean.

### Not touched

The `isolate` tag and its gulp wiring are left in place. Nothing uses the tag any more, so `--exclude-tag isolate` is now a no-op and the tagged-test list could be emptied — but that means touching the parallel and coverage-combining steps in `gulpfile.js`, which is a separate decision and not needed to fix these failures. Happy to follow up if you want the mechanism removed.


cdubz (approved):
👍

PR #1146, opened 2026-09-13
feat: add a notes field to Tummy Time

Daltonganger:
Tummy Time is the only care entry model without a `notes` field:

| Model | `notes` |
| --- | --- |
| BMI | yes |
| DiaperChange | yes |
| Feeding | yes |
| HeadCircumference | yes |
| Height | yes |
| Medication | yes |
| Pumping | yes |
| Sleep | yes |
| Temperature | yes |
| Weight | yes |
| **TummyTime** | **no** |

It has `milestone`, which is a short field for a specific kind of remark, but no
place for anything else. Anyone who wants to record "she seemed tired today" or
"tried the new play mat" alongside a tummy time entry has nowhere to put it,
even though every sibling model offers exactly that.

This adds the field and exposes it everywhere the other models expose theirs:

- `core/models.py` — nullable `TextField`, same as the other ten
- `core/migrations/0038_tummytime_notes.py` — additive, nullable, no default
  (renumbered from 0037 after `0037_alter_diaperchange_color` landed)
- `core/forms.py` — field plus textarea widget, in the advanced fieldset
- `api/serializers.py` — read/write through the API
- `core/timeline.py` — notes are appended to the event details, after
  `milestone`, the same way the other duration models treat theirs
- `openapi-schema.yml` — the `TummyTime` schema block
- `core/templates/core/tummytime_list.html` — details row below the entry
- tests for the timeline, the API write path and the form

The list template uses the same shape as `feeding_list.html` and friends: a
`row-details` row under the entry, rendered only when notes are present. The
colspan correction that was part of the first version of this PR has been
dropped: it landed on `master` separately in c742ffac and 088a7ce4, so this
branch now only inherits it.

### Notes on testing

Full suite green: 429 tests with `--exclude-tag isolate` on top of the current
`master` (6dfc907e), which includes #1134. `black`, `djlint` and `prettier` all
clean. The migration renumbering was checked with `makemigrations --check`,
which reports no missing migrations for this field.

`manage.py makemigrations --check` still reports pre-existing drift on master
for `BMI`, `Feeding`, `HeadCircumference`, `Height` and `Weight` `Meta.ordering`
(introduced in 661f2ff7 without a migration). That is unrelated to this change
and I left it alone; the new field itself is fully covered by 0038.


Daltonganger:
Two things a reviewer should know about `openapi-schema.yml` and migrations, both
pre-existing on `master` and both left alone:

**1. The schema entry is hand-written, because the generator is currently broken.**

I verified this on an unmodified `origin/master` worktree, not just on this branch:

```
$ python manage.py generateschema --title "Baby Buddy API" --file /tmp/schema.yml
AttributeError: 'DjangoFilterBackend' object has no attribute 'get_schema_operation_parameters'
```

That is exactly the breakage [PR #1140](https://github.com/babybuddy/babybuddy/pull/1140)
fixes, which is still open. So `gulp generateschema` cannot run here, and the
committed `openapi-schema.yml` has been stale since v2.10.0.

I added the `notes` property by hand, at the position the serializer puts it
(after `milestone`, before `tags`), using the same shape the other ten models
already use in that file:

```yaml
        notes:
          type: string
          nullable: true
```

Because `prettier` is what formats the generated file in `gulpfile.js`, I ran it
over the result — `prettier . --check` is clean. When #1140 lands, regenerating
should produce this block unchanged; worth a check at that point.

**2. `makemigrations --check` reports drift on `master`, unrelated to this PR.**

```
$ python manage.py makemigrations core --check --dry-run   # clean origin/master
Migrations for 'core':
  ..._alter_bmi_options_alter_feeding_options_and_more.py
    ~ Change Meta options on bmi
    ~ Change Meta options on feeding
    ~ Change Meta options on headcircumference
    ~ Change Meta options on height
    ~ Change Meta options on weight
```

`661f2ff7` changed `Meta.ordering` on those five models without a migration. CI
does not check for this, so it has gone unnoticed. The migration in this PR is
deliberately hand-written to contain only the `notes` field rather than being the
output of `makemigrations`, which would have swept those five unrelated
`AlterModelOptions` operations in with it. The new field itself is fully covered
— `makemigrations --check` no longer mentions `tummytime`.

Daltonganger:
The timeline gap from the review is fixed: `core/timeline._add_tummy_times` now appends the entry's `notes` to the event details, after `milestone`, which is exactly how `_add_sleeps` and the other duration builders already treat theirs. Previously a note saved through the form or API simply never reached the timeline.

There is a regression test for it in `core/tests/test_timeline.py` (`test_tummy_time_notes_appear_in_timeline`) that fails on the first version of this branch and passes now — it creates a Tummy Time entry with both a milestone and a note and asserts both appear in the details of the day's start and end events.

The branch is also rebased on current `master` (6dfc907e): the migration was renumbered to `0038_tummytime_notes` because `0037_alter_diaperchange_color` landed upstream in the meantime, and the colspan correction this PR originally carried has been dropped since c742ffac and 088a7ce4 landed the same fix for all list templates. Full suite green on the new base: 429 tests, `black`/`djlint`/`prettier` clean.

cdubz (approved):
Thanks, @Daltonganger. Looks good!

PR #1147, opened 2026-09-14
fix: empty-state colspan on Notes list (multiple children)

### LINKED ISSUE #918: Incorrect colspan on notes
Like #888 but on `/notes/`:

Screenshot 2024-12-05 at 5 43 31 PM

I'm on v2.7.0.

TianHengZhuang:
## Summary

Fixes the empty-state row on `/notes/` so "No notes found." spans every column when multiple children are shown.

- Table columns: Actions, Time, [Child], Image, Note, Tags
- With `unique_child` false → **6** columns; empty row used `colspan="5"`
- With `unique_child` true → **5** columns (unchanged)

Uses the same `{% if not unique_child %}` pattern as `feeding_list.html`.

## Related

Closes #918

## Testing

- Template-only change
- Visual check: empty notes list with multiple children should show one full-width empty cell


cdubz (approved):
Looks good -- thanks!

PR #1156, opened 2026-09-18
feat: let users complete their own timers

Daltonganger:
## Let users complete the timers they started

Since #1134, converting a timer into an entry requires `core.delete_timer`, because completing a timer deletes it. That also applies to the user who started the timer. A caregiver can start a timer but not turn it into an entry, so the timer stays until someone with deletion permission removes it. Granting `core.delete_timer` avoids that, but it also allows deleting every other user's timers.

This lets the owner of a timer convert it into a feeding, sleep, tummy-time or pumping entry without `core.delete_timer`. Converting another user's timer still requires it. The check is the same in the web interface and the API.

## Keeping ownership meaningful

Owning a timer now carries a right, so ownership should not change as a side effect. Two paths changed it:

- Editing a timer in the web interface assigned it to the user who saved the form.
- A `PATCH` to `/api/timers/<id>/` without a `user` value did the same.

Both now keep the existing owner. Changing the `user` of an existing timer through the API requires `core.delete_timer`, the same permission that allows converting any timer. Creating a timer is unchanged: the user defaults to the requester and can still be set explicitly.

The checks are repeated once the timer row is locked, so a timer reassigned between validation and saving is neither converted by its former owner nor written back with a stale owner.

## Verification

| Check | Result |
| --- | --- |
| `manage.py test --settings=babybuddy.settings.test --exclude-tag isolate` | 436 tests, OK |
| `manage.py test babybuddy.tests.tests_views.ViewsTestCase.test_password_reset --settings=babybuddy.settings.test` (isolate tag) | 1 test, OK |
| `black . --check` | unchanged |
| `djlint --check .` | 0 files would be updated |
| `prettier . --check` | all matched files compliant |

Tests: `core/tests/tests_entry_permissions.py` covers converting an own and another user's timer through the web interface and editing another user's timer. `api/tests.py` covers the same for the API, including a `PATCH` that keeps the owner, a `PATCH` that tries to take over a timer, reassigning a timer with `core.delete_timer`, and a timer reassigned between validation and saving (web and API). The user guide describes the new rule.


Daltonganger:
Both points from the Copilot review are addressed in 3b34202:

- Converting a timer now checks ownership again once the timer row is locked, in the API serializer and in the web forms, so a timer that is reassigned between validation and saving can no longer be converted by its former owner.
- Updating a timer through the API now works on the locked, stored timer, and editing a timer in the web interface only saves the form's fields, so a change of owner made in the meantime is never overwritten with a stale value.

Both cases have tests that reassign the timer between validation and saving. The full suite passes (436 tests plus the isolated one), black, djlint and prettier are clean, and CI is green on Python 3.10 to 3.14.

@cdubz this is ready from my side, feel free to merge it when you're happy with it.

cdubz (approved):

PR #1157, opened 2026-09-18
feat: add an optional access expiry to user accounts

Daltonganger:
## Optional access expiry for user accounts

Access in Baby Buddy lasts until an account is deactivated or deleted. For someone who only needs access for a while, such as a babysitter for a weekend, that means remembering to withdraw it afterwards. #1134 notes this as a limitation of the caregiver role.

This adds an optional **Access expires** time to the user form. Once it has passed, the user:

- is signed out of an existing web session;
- cannot sign in;
- cannot use the API, including with an existing API key.

The account itself is not changed. Moving or clearing the time restores access, so an expiry can be extended without reactivating anything. The value is stored in the user's settings but is not part of the settings form, so users cannot change their own expiry.

## How it is enforced

- Web sessions are checked by a small middleware in its view hook, so users signed in through `REVERSE_PROXY_AUTH` (whose middleware is added after it) are covered as well.
- API tokens are checked by a `TokenAuthentication` subclass in `api/authentication.py`, which is now the configured token authentication class. Session authentication for the API is covered by the middleware.

A migration adds the nullable `access_expires` field to `babybuddy.Settings`.

## Verification

| Check | Result |
| --- | --- |
| `manage.py test --settings=babybuddy.settings.test --exclude-tag isolate` | 434 tests, OK |
| `manage.py test babybuddy.tests.tests_views.ViewsTestCase.test_password_reset --settings=babybuddy.settings.test` (isolate tag) | 1 test, OK |
| `black . --check` | unchanged |
| `djlint --check .` | 0 files would be updated |
| `prettier . --check` | all matched files compliant |

New tests in `babybuddy/tests/tests_access_expiry.py` cover access before expiry, an existing session after expiry, an existing API key after expiry, signing in after expiry, clearing the expiry, setting and clearing it in the user form, and a user trying to change their own expiry through the settings form. The user guide has a short section on it.


Daltonganger:
The Copilot finding is addressed in 4e6a129: an expired account now gets "Your access has expired." from the API instead of "User inactive or deleted.", the same message as the web interface, and the test checks that message. The full suite passes (434 tests plus the isolated one), black, djlint and prettier are clean, and CI is green on Python 3.10 to 3.14.

@cdubz this one is ready from my side as well, feel free to merge it when you're happy with it.

cdubz (approved):

PR #1159, opened 2026-09-20
test: cover same-timestamp timeline order (end before start)

### LINKED ISSUE #928: A timeline-related bug (started activity shown before previously/SAME TIME finished activity)
Hey there!

How are you?

I'm here to report this bug, today.

Let's say my baby just finished taking her breastmilk and automatically slept on the mother's chest. I can't add seconds to my registries when I'm on the phone, so the baby ended their feeding 11:44:00, and started sleeping 11:44:00.

The timeline won't recognize that a just finished activity should be placed before, for cases like this:

image

It's in Portuguese, but one can easily see how, in the timeline, she "slept" before finishing their feed. But actually the time is the very same for both activities (feeding ended 11:44:00, sleeping started 11:44:00). I need to avoid this everytime on the computer, and usually I do, BUT I do think and feel this is still a bug that should be corrected for future versions. If there are two activities with the very same time (seconds included), then it should determine the activity that ended at that timeframe, so it's always shown before the activity that clearly started "after" (even when they are registered with the same time, it's definitely about concepts that faild, so in need to be reinforced - hence I personally classifiy this as a bug).

Best regards and abrazos

TianHengZhuang:
Closes #928 (behavior check)

## Summary

master already sorts same-timestamp timeline events with end before start in core/timeline.py (explicit_type_ordering). This PR adds a regression test so that order does not silently change.

When feeding ends at 11:44:00 and sleep starts at 11:44:00, the end event must appear before the start event on the timeline.

## Testing

- New 	est_same_timestamp_end_before_start in core/tests/test_timeline.py
- Template/production sort left unchanged


TianHengZhuang:
pushed a test fix on this branch

- compare feeding end vs sleep start by model_name instead of raw timestamp equality
- uses formula/bottle like the other timeline tests

CI can re-run whenever

PR #1160, opened 2026-09-21
fix: only show weekly body-measurement stats when span is >= 7 days

### LINKED ISSUE #587: Weight Change in Dashboard is Incorrect

TianHengZhuang:
Closes #587

## Summary

Dashboard change-per-week for weight/height/head circumference/BMI divided the raw delta by (newest-oldest).days/7. Two readings one day apart (2490g -> 2420g) projected a full week and showed about -490.

Per cdubz on #587, do not show a weekly rate when there is less than a week between oldest and newest entries. Same guard on all four measurement stats.

- span < 7 days (or one reading): change_weekly is None; template already renders Not enough data
- span >= 7 days: previous weekly calculation
- avoids division by zero on same-date readings

#589 is an older draft RFC on the same issue and has been idle; this PR implements the narrower hide-until-one-week behavior from the thread.

## Testing

- New test_weight_change_weekly_requires_one_week_span
- Existing fixture-based test_card_statistics still expects the >=7 day fixture values


dashboard/templatetags/cards.py
```diff
@@ -1,738 +1,768 @@
...
-        last_instance = instance
+    if newest == oldest:
+        return stats
 
-    if sleep["count"] > 0:
-        sleep["average"] = sleep["total"] / sleep["count"]
-    if sleep["btwn_count"] > 0:
-        sleep["btwn_average"] = sleep["btwn_total"] / sleep["btwn_count"]
+    days = (newest.date - oldest.date).days
+    if days < 7:
```
pamod-madubashana, line 760: Missing test: same-date entries. Two weight entries on the same date should return None (not 0.0, not a ZeroDivisionError). This was the original latent crash before the span guard. Worth an explicit test case to lock it in.

dashboard/templatetags/cards.py
```diff
@@ -1,738 +1,768 @@
...
+        return stats
```
pamod-madubashana, line 762: Missing test: single entry. With only one measurement, newest == oldest so the if guard skips the computation. Before this PR that returned change_weekly = 0.0; now it returns None. The template handles both, but worth a test to document the intended behavior for one-entry profiles.

pamod-madubashana (changes requested):
The core fix looks correct - just need a couple more test cases to cover the edge cases this introduces.

TianHengZhuang:
pushed the review fixes

- put back @register.inclusion_tag on card_timer_list
- tests for same-date readings and a single reading (both change_weekly None)
- empty check is queryset.exists() now

CI can re-run

